PR #7, opened 2022-01-26
Solution

Nazar-kob:


app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+class DeliveryDrone(FlyingRobot):
+    def __init__(self, name,
+                 weight,
+                 max_weight_load,
+                 current_load,
+                 coords=None):
+        super().__init__(name, weight, coords)
+        self.max_weight_load = max_weight_load
+        self.current_load = current_load\
+            if isinstance(current_load, Cargo)\
```
alexandra-halytska, line 50: I' m not  sure that that lines 50, 51 should be in "__init__"

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+            and current_load.weight < self.max_weight_load else None
+
+    def hook_load(self, cargo):
```
alexandra-halytska, line 53: It's not impossible to check if cargo is instance of Cargo "if isinstance(cargo, Cargo)"

alexandra-halytska: it's beter to remove this line "if isinstance(cargo, Cargo)" from your script

alexandra-halytska (changes requested):

app/main.py
Nazar-kob: Thanks for your comment, but you should understand how, my script works if І remove "if isinstance(cargo, Cargo)" in line 50. Test will not pass, and if I give the argument like None in "Cargo" script will not work correctly!

alexandra-halytska (approved):

app/main.py
```diff
@@ -2,4 +2,56 @@ class Cargo:
...
+
+    def go_left(self, step=1):
+        self.coords[0] -= step
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name, weight, coords=None):
+        super().__init__(name, weight)
+        self.coords: list = [0, 0, 0] if coords is None else\
+            coords if len(coords) > 2 else None
```
MasterpieceElbow, line 32: You don't need code after 1st `else` if you pass coords in super.()__ init__

MasterpieceElbow (changes requested):

app/main.py
Nazar-kob: Thanks, I will change

MasterpieceElbow (approved):

PR #22, opened 2022-01-27
solution

NikitaZH1812:


app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
...
+
+class BaseRobot:
+    def __init__(self, name, weight, coords=None):
+        self.name = name
+        self.weight = weight
+
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
Abnormaltype, line 14: You can write this in one line

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
...
+    def get_info(self):
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name, weight, coords=None):
+        super().__init__(name, weight, coords)
+
+        if coords is None:
+            self.coords = [0, 0, 0]
```
Abnormaltype, line 37: You can write this as: `super().__init__(name, weight, [0, 0, 0] if coords is None else coords)`

Abnormaltype (approved):

app/main.py
MasterpieceElbow: Agree with Alexander

app/main.py
MasterpieceElbow: I think it would be even better `coords if coords else [0, 0, 0,]`

MasterpieceElbow (changes requested):

app/main.py
Abnormaltype: I've approved this because others have the same code and they were approved

IvanRamyk (approved):

PR #23, opened 2022-01-27
'Solution'

Oleksandr563487953:


nikon4uk (approved):

app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
+            name,
+            weight,
+            coords=None
+    ):
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
MasterpieceElbow, line 18: You can make it more beautiful and shorted by writing `self.coords = coords if coords else [0, 0]` Please, write it here and in `FlyingRobot`

MasterpieceElbow (changes requested):

app/main.py
Oleksandr563487953: Yes, i forgot to refactor that. Changes provided!

IvanRamyk (approved):

PR #38, opened 2022-05-23
Solution

lenasharabura:


app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: (int, float),
+        max_load_weight: int,
+        current_load: int,
+        coords: list = None,
```
darten360, line 53: It is unnecessary to add coords here

darten360 (changes requested):

app/main.py
lenasharabura: Really, thx

lenasharabura: i think you are wrong
BaseRobot
- constructor takes name, weight, coords, and saves them

MasterpieceElbow (approved):

PR #39, opened 2022-05-23
Solution

Hectorovich:


app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
     def __init__(self, weight):
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name, weight, coords=None):
```
Aleksei-Isaev, line 7: If I were you, here and below I would make a type annotation. But in this case, please suit yourself.

Aleksei-Isaev (approved):

MasterpieceElbow (approved):

PR #42, opened 2022-05-23
Solution

darten360:


app/main.py
```diff
@@ -2,4 +2,78 @@ class Cargo:
...
+    ):
+        self.name = name
+        self.weight = weight
+        if coords is not None:
+            self.coords = coords
+        else:
+            self.coords = [0, 0]
+
+    def go_forward(self, step: int = 1):
+        new_coords = self.coords
```
MasterpieceElbow, line 21: What is the point to copy `self.coords` to the `new_coords` if you can change `self.coords` in place?

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #46, opened 2022-05-24
Katya's Solution

katevasilenko:


Magus6784 (approved):

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+            current_load: int,
+            coords: list = None,
+    ):
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo):
+        if isinstance(cargo, Cargo):
+            if self.current_load is None \
```
MasterpieceElbow, line 61: Don't use 2-level nested conditions here. Use a single composite `if` statement. Use [pep8 rules](https://peps.python.org/pep-0008/#indentation) to make it correctly.

MasterpieceElbow (changes requested):

PR #49, opened 2022-05-24
Solution

NOSSUDAK:


Y-Havryliv (approved):

app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
...
+    def get_info(self):
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None):
+        if coords is None:
+            coords = [0, 0, 0]
+        else:
+            coords = coords.copy()
```
MasterpieceElbow, line 36: Why do you need to copy `coords`?

app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
...
+            current_load: int,
+            coords: list = None
+    ):
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+        super().__init__(name, weight, coords)
+
+    def hook_load(self, package: Cargo):
+        if self.current_load is None:
+            if package.weight <= self.max_load_weight:
```
MasterpieceElbow, line 61: Here is no point in nested `if`. Write it within a single `if`.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #50, opened 2022-05-24
'Solution'

Yurnerroo:


SergeyDev1996 (approved):
Good job!

app/main.py
```diff
@@ -2,4 +2,56 @@ class Cargo:
     def __init__(self, weight):
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name, weight, coords: list = [0, 0]):
```
MasterpieceElbow, line 7: Don't use a list as a value by default. Because all BaseRobots will use the same `coords` if you will not specify it. Better use by default `None` and assign `coords` in the constructor.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #53, opened 2022-05-24
'Solution'

YuraBeseda:


app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
+            name: str,
+            weight: int,
+            coords: list = None
+    ):
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+
+        if coords is None:
+            self.coords = [0, 0]
```
Chardona, line 18: Lines 15-18 look like repeat of code. Better combine appointment with condition.

app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ):
+        if coords is None:
+            coords = [0, 0, 0]
+        super(FlyingRobot, self).__init__(name, weight, coords)
```
Chardona, line 48: Better replace line 48 right after __init__, because you call second check of "if coords is None" and can re-appoint coords to [0, 0] in FlyingRobot.

YuraBeseda: If super will be right after init, it may appear problems so that coords will be [0, 0] while they must be [0, 0, 0]

P.S I have checked it and there were errors

MasterpieceElbow (approved):

PR #58, opened 2022-05-24
Solution

vasylhnatiuk:


app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
...
+    def go_right(self, step: int = 1):
+        self.coords[0] += step
+
+    def go_left(self, step: int = 1):
+        self.coords[0] -= step
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self, name, weight, coords: list = None):
```
DmytroSvirsa, line 28: use type annotation is a good practice

app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
     def __init__(self, weight):
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+
+    def __init__(self, name, weight, coords: list = None):
```
DmytroSvirsa, line 8: use type annotation is a good practice

app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
...
+
+    def go_back(self, step: int = 1):
+        self.coords[1] -= step
+
+    def go_right(self, step: int = 1):
+        self.coords[0] += step
+
+    def go_left(self, step: int = 1):
+        self.coords[0] -= step
+
```
DmytroSvirsa, line 24: get_info method is missing, yes tests don't need it, but it is in task.

DmytroSvirsa (changes requested):

app/main.py
vasylhnatiuk: thanks)

MasterpieceElbow (approved):

PR #71, opened 2022-06-03
Solution

drivemax26:


tests/test_main.py
```diff
@@ -46,7 +46,7 @@ def test_base_robot_do_not_use_mutable_as_default():
```
IvanRamyk, line 46: You shouldn't change tests :)

IvanRamyk (changes requested):
Please, undo changes in the tests

IvanRamyk (approved):
Good job!

PR #82, opened 2022-07-08
Solution

denlubn:


app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name, weight, coords: list = None):
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
Abnormaltype, line 11: you can write this in one line

app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
...
+        self.coords[0] -= step
+
+    def get_info(self):
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name, weight, coords: list = None):
+        if coords is None:
+            coords = [0, 0, 0]
```
Abnormaltype, line 34: the same here

Abnormaltype (changes requested):

app/main.py
denlubn: image
image
image

denlubn: Тут не получилось

Abnormaltype (changes requested):
try to write like this `self.coords = [0, 0] if coords is None else coords`

PR #107, opened 2022-08-18
Solution

KatyaVasylieva:


DmytroSvirsa (changes requested):
Well done, but let`s use a type annotation not only for `coords: list`

DmytroSvirsa (approved):
GJ!

PR #128, opened 2022-09-16
Solution

sbublinskyi:


mltdll (approved):
Good job.

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
     def __init__(self, weight):
```

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None):
```
Nattalli, line 7: `list[int]` should be here

Nattalli (changes requested):

app/main.py
```diff
@@ -1,5 +1,73 @@
 class Cargo:
-    def __init__(self, weight):
+    def __init__(self, weight) -> None:
```
Nattalli, line 2: add type annotation for `weight` variable too

Nattalli (changes requested):

Nattalli (approved):
Well done! 👍🏻

PR #131, opened 2022-09-16
Completed

maxitoska:


app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 weight,
+                 name,
+                 max_load_weight,
+                 current_load,
+                 coords=None
+                 ):
```
IvanRamyk, line 51:
```suggestion
    def __init__(
        self,
        weight,
        name,
        max_load_weight,
        current_load,
        coords=None
 ):
```

IvanRamyk: I prefer to define parameters in this way. But it's just a suggestion

IvanRamyk (approved):

app/main.py
maxitoska: ok , i will use it in the next tasks , thx

PR #137, opened 2022-09-19
'Solution'

AVyha:


app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
     def __init__(self, weight):
         self.weight = weight
 
-# write your code here
+
+class BaseRobot(Cargo):
```
wQuelS, line 6: Does BaseRobot class really need to inherite from Cargo?

wQuelS (changes requested):

app/main.py
AVyha: i think really need because in readme write something like this
image

wQuelS: Ok, i guess task is not that clear. Mentor will answer that question anyway:)

wQuelS (approved):

app/main.py
AVyha: Thanks for the review

Nattalli: @wQuelS is right. That class should not be a child of `Cargo`. The task is to create 3 classes (it is their general count) and use the inheritance (but it doesn't mean that every class should use the inheritance)

app/main.py
```diff
@@ -1,5 +1,59 @@
 class Cargo:
-    def __init__(self, weight):
+    def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot(Cargo):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```

app/main.py
```diff
@@ -1,5 +1,59 @@
...
+
+    def get_info(self) -> str:
+        return f'Robot: {self.name}, Weight: {self.weight}'
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None) -> None:
```
Nattalli, line 32: it's better to use `list[int]` here

app/main.py
```diff
@@ -1,5 +1,59 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: Cargo = None,
+                 coords: list = None) -> None:
```
Nattalli, line 49: it's better to use `list[int]` here

Nattalli (changes requested):

Nattalli (approved):
Well done! 👍🏻

PR #139, opened 2022-09-20
Solution

StasOkhrym:


app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+    def go_down(self, step: int = 1) -> None:
+        self.coords[-1] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int
```
IvanStored, line 53: why DeliveryDrone doesn't accept `coords` argument?

IvanStored (changes requested):

IvanStored (approved):

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
     def __init__(self, weight):
```
Nattalli, line 2: add type anotation

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None):
```
Nattalli, line 10: it's better to use `list[int]` here

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None):
```
Nattalli, line 35: it's better to use `list[int]` here

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+        self.coords[-1] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int,
+                 coords: list = None):
```
Nattalli, line 52: it's better to use `list[int]` here

Nattalli (changes requested):

Nattalli (approved):
Good job!

PR #140, opened 2022-09-20
Solution for the task

olia-trofymchuk:


app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        super(FlyingRobot, self).__init__(name=name,
```
wQuelS, line 34: in python 3.10 u don't have to put anything in super brackets.

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
...
+
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self, name: str, weight: int, coords: list = None,
+                 max_load_weight: int = 0,
+                 current_load: Cargo = None) -> None:
+        super(FlyingRobot, self).__init__(name=name,
```
wQuelS, line 49: same

wQuelS (changes requested):
Everything else looks great, good job

app/main.py
olia-trofymchuk: Thx for your review. I will notice it!

app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
     def __init__(self, weight):
```

app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```

app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Nattalli, line 31: it's better to use `list[int]` here

app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
...
+
+    def go_up(self, step: int = 1) -> None:
+        self.coords[2] += step
+
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self, name: str, weight: int, coords: list = None,
```
Nattalli, line 44: it's better to use `list[int]` here

Nattalli (changes requested):

app/main.py
```diff
@@ -1,5 +1,61 @@
 class Cargo:
-    def __init__(self, weight):
+    def __init__(self, weight) -> None:
```
Nattalli, line 2: add type annotation for `weight` variable

Nattalli (changes requested):

Nattalli (approved):
Excellent! 🌞

PR #145, opened 2022-09-22
'Solution'

hasslieb3:


app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None):
+        if coords is None:
+            coords = [0, 0]
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+
+    def go_forward(self, step: int = 1):
```
AVyha, line 14: add annotation for function like this
`def go_forward(self, step: int = 1) -> None:`
use this annotation to all function

AVyha (changes requested):
use function annotation to all function

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
     def __init__(self, weight):
```

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None):
```

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self, name: str, weight: int,
+            max_load_weight: int, current_load: Cargo,
+            coords: list = None):
```
Nattalli, line 47: and here

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None):
```
Nattalli, line 31: and here `list[int]`

Nattalli, line 31: and here `list[int]`

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self, name: str, weight: int,
+            max_load_weight: int, current_load: Cargo,
+            coords: list = None):
+        if coords is None:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self. current_load = current_load
```
Nattalli, line 52: does this line works with empty space between `self.` and `current_load`?

Nattalli (changes requested):

MasterpieceElbow (approved):
GJ!

PR #147, opened 2022-09-22
Solution

shurushku:


IvanStored (approved):
seems good

app/main.py
```diff
@@ -1,5 +1,58 @@
 class Cargo:
-    def __init__(self, weight):
+    def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot(Cargo):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Nattalli, line 7: `list[int]` has to be here

app/main.py
```diff
@@ -1,5 +1,58 @@
 class Cargo:
-    def __init__(self, weight):
+    def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot(Cargo):
```
Nattalli, line 6: oh no, it should not be a child of `Cargo` class

Nattalli: `Cargo` needed for `DeliveryDrone` only

app/main.py
```diff
@@ -1,5 +1,58 @@
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Nattalli, line 29: `list[int]` has to be here

app/main.py
```diff
@@ -1,5 +1,58 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            coords: list = None,
```
Nattalli, line 46: `list[int]` has to be here

Nattalli (changes requested):

Nattalli (approved):
Excellent! 🌞

PR #151, opened 2022-09-22
Solution)

Gr1zzZzzly:


mltdll (approved):

app/main.py
```diff
@@ -1,5 +1,57 @@
 class Cargo:
-    def __init__(self, weight):
+    def __init__(self, weight: int):
```

app/main.py
```diff
@@ -1,5 +1,57 @@
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Nattalli, line 7: `list[int]` has to be here

app/main.py
```diff
@@ -1,5 +1,57 @@
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self):
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Nattalli, line 29: `list[int]` has to be here

app/main.py
```diff
@@ -1,5 +1,57 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: Cargo,
+                 coords: list = None
```
Nattalli, line 46: `list[int]` has to be here

Nattalli (changes requested):

Nattalli (approved):
Good job!

PR #161, opened 2022-09-24
Solution

linaSam:


app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self, name: str, weight: int, coords: list[int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
```
wQuelS, line 12: you can do this in more "pythonic" way, in one line. Like this:
self.coords = ... if coords is None else ..

app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self, name: str, weight: int, coords: list[int] = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = coords
```
wQuelS, line 38: same here. You can shorten 3 lines ( 37-40) into one, by doing onliner with ternary operator

wQuelS (changes requested):
Everything else looks really good to me:)

app/main.py
linaSam: Thanks for idea!

wQuelS (approved):
Seems good to me :)

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
     def __init__(self, weight):
```

Nattalli (changes requested):

app/main.py
```diff
@@ -1,5 +1,62 @@
 class Cargo:
-    def __init__(self, weight):
+    def __init__(self, weight) -> None:
```
Nattalli, line 2: add type annotation for the `weight` variable too

Nattalli (approved):
Excellent! 🌞

PR #164, opened 2022-09-25
Solution

luxuriant777:


app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
     def __init__(self, weight):
```

Nattalli (changes requested):

app/main.py
luxuriant777: added

TPolina (approved):
Well done!

PR #165, opened 2022-09-25
'Solution'

serviza:


FranchukS (approved):
Good!

serviza:
> Good!

Thanks!)

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
     def __init__(self, weight):
```
Nattalli, line 2: add type annotations

Nattalli (changes requested):

app/main.py
serviza: Done!

Nattalli (approved):
Good! 👍🏻

PR #166, opened 2022-09-25
Solution

FranchukS:


HStakhiv (approved):
Good!

app/main.py
```diff
@@ -1,5 +1,58 @@
 class Cargo:
-    def __init__(self, weight):
+    def __init__(self, weight) -> None:
```
Nattalli, line 2: add type annotation for `weight`

Nattalli (changes requested):

app/main.py
FranchukS: type annotation added

Nattalli (approved):
Well done!

PR #168, opened 2022-09-25
Solution

HStakhiv:


FranchukS (approved):
GJ!

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
     def __init__(self, weight):
```

Nattalli (changes requested):

app/main.py
HStakhiv: Done

Nattalli (approved):
Well done!

PR #172, opened 2022-09-26
Solution

skrypka-dmytro:


app/main.py
```diff
@@ -1,5 +1,69 @@
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
```
ketstap162, line 13: if not coords

ketstap162 (changes requested):

app/main.py
```diff
@@ -1,5 +1,69 @@
...
+                 max_load_weight: int,
+                 current_load: Cargo,
+                 coords: list[int] = None) -> None:
+        super().__init__(name, weight, coords)
+
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
```
ketstap162, line 65: same

ketstap162 (changes requested):

ketstap162 (approved):

app/main.py
```diff
@@ -1,5 +1,69 @@
 class Cargo:
-    def __init__(self, weight):
+    def __init__(self, weight: int):
```

Nattalli (changes requested):

app/main.py
Nattalli: yes, these commentaries are not required but it will be better for you as it's more pythonic

Nattalli (approved):

PR #178, opened 2022-09-26
solve the task

Cerne13:


Vasyl-Poremchuk (approved):
GJ!

app/main.py
```diff
@@ -1,5 +1,70 @@
 class Cargo:
-    def __init__(self, weight):
+    def __init__(self, weight: int):
```

app/main.py
```diff
@@ -1,5 +1,70 @@
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords:
+            list[int] = None
```
Nattalli, line 12: can it be in one line?

app/main.py
```diff
@@ -1,5 +1,70 @@
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords:
+            list[int] = None
```
Nattalli, line 40: and here

Nattalli (changes requested):

app/main.py
Cerne13: Thank you!

Nattalli (approved):
Good job! 😸

PR #179, opened 2022-09-26
Solution

Qusert17:


Vasyl-Poremchuk (approved):
Not bad!

app/main.py
```diff
@@ -2,4 +2,71 @@ class Cargo:
     def __init__(self, weight):
```

Nattalli (changes requested):

app/main.py
Qusert17: Done)

Nattalli (approved):
👍🏻

PR #188, opened 2022-09-26
Develop

pazyn404:
I modify tests cause original doesn't even run. I just delete block for another version, so it's not cheating)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,5 +1,74 @@
+import copy
+
+
 class Cargo:
     def __init__(self, weight):
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+
+        self.name = name
+        self.weight = weight
+        if coords:
+            self.coords = coords
+        else:
+            self.coords = [0, 0]
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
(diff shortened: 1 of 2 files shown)


app/main.py
```diff
@@ -1,5 +1,73 @@
...
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+
+        if coords:
+            super().__init__(name, weight, coords)
+        else:
+            super().__init__(name, weight, [0, 0, 0])
+
```
StasOkhrym, line 48: Here as well
but it's up to you

app/main.py
```diff
@@ -1,5 +1,73 @@
...
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+
+        self.name = name
+        self.weight = weight
+        if coords:
+            self.coords = coords
+        else:
+            self.coords = [0, 0]
```
StasOkhrym, line 20: You could implement coord initialization it a bit easier

app/main.py
```diff
@@ -1,5 +1,73 @@
...
+                 coords: list[int] = None) -> None:
+
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if not self.current_load and cargo.weight <= self.max_load_weight:
+            self.current_load = copy.deepcopy(cargo)
+
```
StasOkhrym, line 71: no need to use deepcopy, imho.

StasOkhrym (approved):

Nattalli (changes requested):

Nattalli (approved):
Awesome! 😎

PR #205, opened 2022-10-09
Solution of py-delivery-drone

YLobunets:


app/robot.py
```diff
@@ -0,0 +1,58 @@
+from __future__ import annotations
```
IvanRamyk, line 1: write the code in the main module. 
you can keep classes declarations and usage tests in one module

IvanRamyk (changes requested):

app/robot.py
YLobunets: Everything is transferred to main

IvanRamyk (approved):
Good job!

PR #227, opened 2022-11-08
Done

DevSpaciX:


app/main.py
```diff
@@ -1,5 +1,60 @@
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self, name: str, weight: int, coords: None = None) -> None:
+        super().__init__(name, weight, coords)
+        if self.coords == [0, 0]:
+            self.coords = [0, 0, 0]
```
DmytroSvirsa, line 36: better do not provide coords into super().__init__ and check for None it makes your code mo flrxible to changes id BaseRobot class

DmytroSvirsa (changes requested):

app/main.py
DevSpaciX: Thanks ! recomitted

DmytroSvirsa (approved):

PR #229, opened 2022-11-09
Solution

tinabushuieva:


app/main.py
```diff
@@ -1,5 +1,65 @@
...
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self, name: str,
+                 weight: float,
+                 coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = coords
+        if self.coords is None:
+            self.coords = [0, 0, 0]
```
DmytroSvirsa, line 40: you don`t need coords in super().__init__()

DmytroSvirsa (changes requested):

PR #251, opened 2022-12-04
solution

kleotan901:


app/main.py
```diff
@@ -1,5 +1,66 @@
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+        if coords is None:
+            self.coords = [0, 0]
+
```
OleksandrPozden, line 13: re-write it in one line.

app/main.py
```diff
@@ -1,5 +1,66 @@
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
```
OleksandrPozden, line 38: here as well

app/main.py
```diff
@@ -1,5 +1,66 @@
...
+        self.coords[0] -= step
+        return self.coords
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
```
OleksandrPozden, line 36: Also, you do need to pass here coords as a parameter.

OleksandrPozden (changes requested):

app/main.py
Serhii-Leonenko: I think that Oleksandr wants to say that you don't need to pass coords in super, remove. And remove 36 line `self.coords = coords` it's redundant also.

Serhii-Leonenko (changes requested):

YuriiKindrat (approved):

PR #259, opened 2022-12-11
Solution

andrii-skryl:


app/main.py
```diff
@@ -1,5 +1,61 @@
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+        if coords is None:
+            self.coords = [0, 0]
```
YuriiKindrat, line 12: u can write it in one line: self.coords = coords or [0,0]

YuriiKindrat (changes requested):

PR #268, opened 2022-12-25
1st try solution

AnnaKabatova:


app/main.py
```diff
@@ -1,5 +1,65 @@
...
+        self.weight = weight
+
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
         self.weight = weight
+        self.coords = coords
+        if self.coords is None:
+            self.coords = [0, 0]
```
OleksandrPozden, line 15:
```suggestion
        self.coords = coords or [0,0]
```
apply the same for another case

app/main.py
```diff
@@ -1,5 +1,65 @@
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        if self.coords is None:
+            self.coords = [0, 0, 0]
+        elif len(self.coords) == 2:
+            self.coords.append(0)
```
OleksandrPozden, line 39: this is redundant

OleksandrPozden (changes requested):

app/main.py
AnnaKabatova: it works, but... how?

andrewbelo (approved):
GJ

PR #285, opened 2023-01-07
Solution

Phoenix-Erazer:


app/main.py
```diff
@@ -1,5 +1,57 @@
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 coords: list = None,
+                 current_load: "Cargo" = None
+                 ) -> None:
```
Vasyl-Smutok, line 47: it will look better that way
```
    def __init__(
        self,
        name: str,
        weight: int,
        max_load_weight: int,
        coords: list = None,
        current_load: "Cargo" = None,
    ) -> None:
```

Vasyl-Smutok (changes requested):

Vasyl-Smutok (approved):
Well Done 👍

PR #290, opened 2023-01-13
solved task

SDobzhynetskyi:


app/main.py
```diff
@@ -1,5 +1,81 @@
...
+        self.name = name
+        self.weight = weight
+        self.coords = [0, 0] if coords is None else coords
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+    def go_forward(self, step: int = 1) -> list:
+        one_step_move = self.coords
+        one_step_move[1] += 1 if step is None else step
```
OleksandrPozden, line 22:
```suggestion
        one_step_move[1] += step or 1
```
check this out

app/main.py
```diff
@@ -1,5 +1,81 @@
...
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: float,
+            coords: list = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = [0, 0] if coords is None else coords
```
OleksandrPozden, line 15:
```suggestion
        self.coords = coords or [0,0]
```
Apply this for Flying robot as well

app/main.py
OleksandrPozden, line 22: By the way. As you already have a default argument value of 1. So It is gonna be convenient to manipulate with the value step. Like this
`one_step_move[1] += step`
It is gonna work out for you.
Another thing to comment on that is instead of creating a new variable that actually is a copy of self.coords, you can directly access to it like this 
`self.coords[1] += step`
`return self.coords`

app/main.py
```diff
@@ -1,5 +1,81 @@
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+    def go_forward(self, step: int = 1) -> list:
+        one_step_move = self.coords
+        one_step_move[1] += 1 if step is None else step
+        return one_step_move
+
+    def go_back(self, step: int = 1) -> list:
```
OleksandrPozden, line 25:
```suggestion
    def go_back(self, step: int = 1) -> list[int]:
```
use more descriptive type annotation

OleksandrPozden (changes requested):

app/main.py
```diff
@@ -1,5 +1,81 @@
...
+        one_step_move[1] += step
+        return one_step_move
+
+    def go_back(self, step: int = 1) -> list[int]:
+        one_step_move = self.coords
+        one_step_move[1] -= step
+        return one_step_move
+
+    def go_right(self, step: int = 1) -> list[int]:
+        one_step_move = self.coords
```
andrewbelo, line 31: Ці функції немаю нічого повертати, просто додай/відніми і на тому
Не треба зайвих змінних

andrewbelo: Мається на увазі логіка ``inplace``

andrewbelo (changes requested):

PR #302, opened 2023-01-26
Solution

ilushasm:


app/main.py
```diff
@@ -1,5 +1,71 @@
...
+    def __init__(
+            self, name:
+            str, weight: int,
+            coords: list[int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+        if coords is None:
+            self.coords = [0, 0]
```
ArsenPidhoretskyi, line 16: use inline if

ArsenPidhoretskyi (approved):
Good job. Also, check comment

PR #318, opened 2023-02-05
Solution

bohdan-yatsyna:


app/main.py
```diff
@@ -1,5 +1,65 @@
...
 class Cargo:
-    def __init__(self, weight):
+    def __init__(self, weight: int | float) -> None:
+        self.weight = weight
+
+
+class BaseRobot:
+    def __init__(
+            self, name: str, weight: int, coords: Optional[list] = None
+    ) -> None:
```
Vasyl-Smutok, line 12:
```suggestion
def __init__(
        self, 
        name: str, 
        weight: int, 
        coords: Optional[list] = None
) -> None:
```

app/main.py
```diff
@@ -1,5 +1,65 @@
...
+        self.name = name
         self.weight = weight
+        self.coords = [0, 0] if coords is None else coords
```
Vasyl-Smutok, line 15: also you can write it like this 
```
```

app/main.py
```diff
@@ -1,5 +1,65 @@
...
+        self.coords[0] -= step_left
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self, name: str, weight: int, coords: Optional[list] = None
+    ) -> None:
```
Vasyl-Smutok, line 36:
```suggestion
def __init__(
        self, 
        name: str, 
        weight: int, 
        coords: Optional[list] = None
) -> None:
```

Vasyl-Smutok (changes requested):

app/main.py
bohdan-yatsyna: Nice hint, thank you. All fixed.

Vasyl-Smutok (approved):
Well Done 👍

PR #322, opened 2023-02-07
Solution

AlimFreight:


app/main.py
```diff
@@ -1,5 +1,72 @@
...
+                 name: str,
+                 weight: int,
+                 coords: list = None
+                 ) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
katevasilenko, line 17: Try to write it in one line

app/main.py
```diff
@@ -1,5 +1,72 @@
...
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None
+                 ) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
+        else:
+            self.coords = coords
```
katevasilenko, line 45: Try to write it in one line

app/main.py
```diff
@@ -1,5 +1,72 @@
...
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int | Cargo = None,
```
katevasilenko, line 59: The type is only Cargo, I believe

app/main.py
```diff
@@ -1,5 +1,72 @@
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None
+                 ) -> None:
```
katevasilenko, line 11:
```suggestion

    def __init__(
             self, 
             name: str, 
             weight: int,
             coords: list = None
     ) -> None:
```
It would be great if you provide this style with parentheses all over the project

katevasilenko (changes requested):

app/main.py
AlimFreight: If I do it las ternary operator "self.coords = [0, 0] if coords is None else self.coords = coords" 
I'm getting this error app/main.py:15:10: E999 SyntaxError: cannot assign to attribute here. Maybe you meant '==' instead of '='?

app/main.py
```diff
@@ -1,5 +1,67 @@
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
```
Vasyl-Smutok, line 11:
```suggestion
            coords: list[int] = None
```

app/main.py
```diff
@@ -1,5 +1,67 @@
...
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = [0, 0] if coords is None else coords
```
Vasyl-Smutok, line 15: this option is also possible)

```suggestion
        self.coords = coords or [0, 0]
```

Vasyl-Smutok (approved):
Well Done 👍 
Also check comment

PR #325, opened 2023-02-08
Solution

SafonovVladimir:
add my Solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -57,7 +57,7 @@ class DeliveryDrone(FlyingRobot):
             name: str,
             weight: int,
             max_load_weight: int,
-            current_load: int | None,
+            current_load: Cargo = None,
             coords: list = None
     ) -> None:
         super().__init__(name, weight, coords=coords)


app/main.py
```diff
@@ -1,5 +1,63 @@
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords:
+            self.coords = coords
+        else:
+            self.coords = [0, 0]
```
AndereLion, line 13: U can use one line ternary operator here .

app/main.py
```diff
@@ -1,5 +1,63 @@
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight)
+        if coords:
+            self.coords = coords
+        else:
+            self.coords = [0, 0, 0]
```
AndereLion, line 37: Same)

AndereLion (changes requested):

app/main.py
SafonovVladimir: Done

AndereLion (approved):

app/main.py
```diff
@@ -1,5 +1,57 @@
...
+    def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords else [0, 0]
```
danylott, line 10: this solution will work here, because it’s forbidden to have [] coords, but in other situation, if you check just if coords, but not if coords is None - then you won’t be able to specify empty array into coords at all
>>> bool([])
False

app/main.py
```diff
@@ -1,5 +1,57 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: (int, None),
```
danylott, line 46: Not correct type annotations. use `|`

app/main.py
```diff
@@ -1,5 +1,57 @@
...
+                 coords: list = None) -> None:
```
danylott, line 47: Provide parameters on new line as black suggests (code style):
image

danylott (changes requested):

app/main.py
SafonovVladimir: Fixed

app/main.py
SafonovVladimir: Fixed

app/main.py
SafonovVladimir: I don't understand, won't add [0, 0] if an empty array ([]) comes in?

SafonovVladimir: Fixed

app/main.py
```diff
@@ -1,5 +1,72 @@
...
+
+
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: int | None,
```
wQuelS, line 60: are u sure `current_load` is `int` here ?

wQuelS (changes requested):

app/main.py
SafonovVladimir: Ah, no. It's a Cargo and default = None
Fixed

wQuelS (approved):
Nice job

PR #327, opened 2023-02-08
1 commit

AndereLion:


MaybeTI (approved):
GJ!!!

app/main.py
```diff
@@ -1,5 +1,65 @@
...
+    def __init__(self, weight: int) -> None:
+        self.weight = weight
+
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+
+        self.name = name
         self.weight = weight
+        self.coords = coords if coords else [0, 0]
```
danylott, line 14: this solution will work here, because it’s forbidden to have [] coords, but in other situation, if you check just if coords, but not if coords is None - then you won’t be able to specify empty array into coords at all
>>> bool([])
False

app/main.py
```diff
@@ -1,5 +1,65 @@
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None) -> None:
```
danylott, line 36: It's better to put parameters as `black` suggests:
image

app/main.py
```diff
@@ -1,5 +1,65 @@
...
+                 weight: int,
+                 coords: list = None,
+                 max_load_weight: int = None,
+                 current_load: int = None) -> None:
+
+        self.coords = coords if coords else [0, 0, 0]
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+        super().__init__(name, weight, self.coords)
+
```
danylott, line 59: It's a bad practice to call __init__ at the end, but not at the beginning

danylott (changes requested):

app/main.py
AndereLion: Fixed

app/main.py
AndereLion: Fixed

app/main.py
AndereLion: Fixed

app/main.py
```diff
@@ -1,5 +1,66 @@
...
-    def __init__(self, weight):
+    def __init__(self, weight: int) -> None:
+        self.weight = weight
+
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
         self.weight = weight
+        self.coords = coords if type(coords) == list else [0, 0]
```
wQuelS, line 13: check `if coord is None` no need type check

app/main.py
```diff
@@ -1,5 +1,66 @@
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        self.coords = coords if coords else [0, 0, 0]
```
wQuelS, line 38: Read comment from Dania. It is better to use `if coords is not None`

app/main.py
```diff
@@ -1,5 +1,66 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self, name: str,
+            weight: int,
+            coords: list = None,
+            max_load_weight: int = None,
+            current_load: int = None
```
wQuelS, line 54: are u sure `current_load` is `int` here ?

app/main.py
```diff
@@ -1,5 +1,66 @@
...
+    ) -> None:
+        self.coords = coords if type(coords) == list else [0, 0, 0]
```
wQuelS, line 56: same, u don;t need type check

app/main.py
```diff
@@ -1,5 +1,66 @@
...
+        super().__init__(name, weight, self.coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, other: Cargo) -> None:
```
wQuelS, line 61: `other` is bad name here

wQuelS (changes requested):

app/main.py
AndereLion: Fixed

app/main.py
AndereLion: Fixed

app/main.py
AndereLion: Done.

app/main.py
AndereLion: Same

app/main.py
AndereLion: Fixed

app/main.py
```diff
@@ -1,5 +1,66 @@
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        self.coords = coords if coords is not None else [0, 0, 0]
```
YuriiKindrat, line 38: no need to create attribute before super func, just make a check of coords inside super func

app/main.py
YuriiKindrat: no need to check coords here, it is already checked in parent class

YuriiKindrat (changes requested):

app/main.py
AndereLion: Fixed.

YuriiKindrat (changes requested):

app/main.py
AndereLion: Fixed 38 line

YuriiKindrat (approved):

PR #328, opened 2023-02-09
Solution

Roman28101:


app/main.py
```diff
@@ -1,5 +1,67 @@
...
+    def __init__(self, weight: int) -> None:
+        self.weight = weight
+
+
+class BaseRobot:
+
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
         self.weight = weight
+        self.coords = coords
```
vasylhnatiuk, line 14: would be better to write `self.coords = coords if coords is not None else [0, 0, 0]`

app/main.py
```diff
@@ -1,5 +1,67 @@
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
```
vasylhnatiuk, line 38: the same here

app/main.py
```diff
@@ -1,5 +1,67 @@
...
+
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Union[None, Cargo],
+            coords: list = None
```
vasylhnatiuk, line 56: would be great to add `list[int]`

vasylhnatiuk (approved):

PR #329, opened 2023-02-09
Solution

NEkropo1:


app/main.py
```diff
@@ -1,5 +1,71 @@
...
+    def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
```
DavidGlusch, line 10: coords = [3] is incorrect input, so list[int] is incorrect atribute annotation

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
```
DavidGlusch, line 38: same

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 coords: list[int] = None,
```
DavidGlusch, line 56: same

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+                 current_load: Cargo = None) -> None:
```
DavidGlusch, line 57: can also be None

DavidGlusch (changes requested):

app/main.py
NEkropo1: Agree here

app/main.py
NEkropo1: int is about type of attributes inside list, no?
I mean:
image
but this is ok:
image
So even if I implement list[int, int] as i suppose you expected me to do, result will be still ok:
image

Not arguing, just want to know the truth )

NEkropo1: P.S. type of VARIABLES inside list

image

app/main.py
NEkropo1: So, for example, if we expect by default that value(X) to be None, and change some behavior according to if we implement any optional attribute, it's good idea to mention it's X | None, but when we by default expect value, and by making optional None, create None by default, it's just complex.

DavidGlusch (approved):

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+
+        self.name = name
+        self.weight = weight
+        if not coords:
+            coords = [0, 0]
```
danylott, line 15: this solution will work here, because it’s forbidden to have [] coords, but in other situation, if you check just if coords, but not if coords is None - then you won’t be able to specify empty array into coords at all
>>> bool([])
False

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 coords: list[int] = None,
+                 current_load: Cargo = None) -> None:
+
+        super().__init__(name=name,
+                         weight=weight,
+                         coords=coords if coords else [0, 0, 0])
```
danylott, line 61: Same here about new lines during call

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if not self.current_load and cargo.weight <= self.max_load_weight:
```
danylott, line 67: Here also it's better to check for None

danylott (changes requested):

app/main.py
NEkropo1: Yawni) If u could've watch changes in my code, you would've seen that I removed that check

app/main.py
NEkropo1: image

app/main.py
NEkropo1: Still can't get why you need to specify empty array into coords, you gonna need to rewrite move methods, in case someone specifies an empty array and forgets to specify it commonly or 21391023 other cases. I thought overall idea is yagni kiss dry, readme, checkilst and doing task.

wQuelS (approved):
Good

PR #330, opened 2023-02-09
Solution

DavidGlusch:


app/main.py
```diff
@@ -1,5 +1,67 @@
...
         self.weight = weight
 
+
 # write your code here
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int, int] = None) -> None:
```
ZabFTFT, line 12: I am not sure and I cant find this straight rule of PEP but one of mentors pushed me to move ")-> None" to the next line of def level. But I won't reject it because mb your style is appropriate as well.
Well done! I also liked your 1 line if else solution inside init.

ZabFTFT (approved):

app/main.py
YuriiKindrat: @ZabFTFT is right
image
this rule also works here )

app/main.py
```diff
@@ -1,5 +1,67 @@
...
+        self.coords[0] -= step
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int, int, int] = None) -> None:
+
+        coords = coords if coords else [0, 0, 0]
```
YuriiKindrat, line 39: u can write this condition inside super.init method

app/main.py
```diff
@@ -1,5 +1,67 @@
...
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: None | Cargo,
+                 coords: list[int, int, int] = None) -> None:
+
+        coords = coords if coords else [0, 0, 0]
```
YuriiKindrat, line 57: the same here

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,5 +1,81 @@
...
+
+    def unhook_load(self) -> None:
+        self.current_load = None
+
+
+drone1 = DeliveryDrone(name="Jim",
+                       weight=18,
+                       coords=[1],
+                       max_load_weight=20,
+                       current_load=None)
```
YuriiKindrat, line 81: remove unnecessary code here

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,5 +1,74 @@
...
+            current_load: None | Cargo,
+            coords: list[int, int, int] = None
+    ) -> None:
+
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+        super().__init__(
+            name,
+            weight,
+            coords=coords if coords else [0, 0, 0]
```
YuriiKindrat, line 66: no need to check coords here, they already checked in parent class

YuriiKindrat (changes requested):

PR #332, opened 2023-02-10
Solution

Zenionoi:


Maytreyaya (approved):
GJ

app/main.py
```diff
@@ -1,5 +1,66 @@
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int | float] = None) -> None:
```
danylott, line 11: Please, provide parameters like `black` suggests:
image

app/main.py
```diff
@@ -1,5 +1,66 @@
...
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords else [0, 0]
```
danylott, line 14: this solution will work here, because it’s forbidden to have [] coords, but in other situation, if you check just if coords, but not if coords is None - then you won’t be able to specify empty array into coords at all
>>> bool([])
False

app/main.py
```diff
@@ -1,5 +1,66 @@
...
+                 weight: int,
+                 coords: list[int | float] = None,
+                 max_load_weight: int = 0,
+                 current_load: Cargo = None) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if not self.current_load and cargo.weight <= self.max_load_weight:
```
danylott, line 62: Here also you can check for None

danylott (changes requested):

wQuelS (approved):
GJ

PR #333, opened 2023-02-10
Robots Solution

Maytreyaya:


app/main.py
```diff
@@ -1,5 +1,73 @@
+from __future__ import annotations
+
+
 class Cargo:
-    def __init__(self, weight):
+    def __init__(self, weight: int) -> Cargo:
```
kuznetsb, line 5: __init__ here doesn't return Cargo object

app/main.py
```diff
@@ -1,5 +1,73 @@
...
+        self.weight = weight
+
+
+class BaseRobot(Cargo):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None) -> BaseRobot:
```
kuznetsb, line 13: same as for Cargo and you can specify list[int] in type annotation for coords

app/main.py
```diff
@@ -1,5 +1,73 @@
...
+        self.name = name
         self.weight = weight
+        self.coords = coords
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
kuznetsb, line 20: It seems better firstly to determine are coords [0, 0] or there are coords and then write them to instance of class : self.coords = coords

app/main.py
```diff
@@ -1,5 +1,73 @@
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None) -> FlyingRobot:
```
kuznetsb, line 42: same return Annotation issue

app/main.py
```diff
@@ -1,5 +1,73 @@
...
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
+        else:
+            self.coords = coords
```
kuznetsb, line 47: Same, I'd better check for coords and then pass them to BaseRobot instance by super, in that case you don't need to do self.coords in FlyingRobot __init__ at all

app/main.py
```diff
@@ -1,5 +1,73 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int,
+                 coords: list = None) -> DeliveryDrone:
```
kuznetsb, line 62: -> None ))

kuznetsb (changes requested):
Check return annotation for init ))

app/main.py
```diff
@@ -1,5 +1,72 @@
...
-    def __init__(self, weight):
+    def __init__(self, weight: int) -> None:
+        self.weight = weight
+
+
+class BaseRobot(Cargo):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
```
danylott, line 13: Use `black` codestyle:
image

app/main.py
```diff
@@ -1,5 +1,72 @@
...
+        self.name = name
         self.weight = weight
+        if coords:
+            self.coords = coords
+        else:
+            self.coords = [0, 0]
+
```
danylott, line 20: You don't need this else  - just check if coords is None - then coords = [0, 0], and then self.coords = coords

app/main.py
```diff
@@ -1,5 +1,72 @@
...
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        super().__init__(name, weight)
+        if coords:
+            self.coords = coords
+        else:
+            self.coords = [0, 0, 0]
```
danylott, line 46: Same here

app/main.py
```diff
@@ -1,5 +1,72 @@
...
+                 max_load_weight: int,
+                 current_load: int,
+                 coords: list[int] = None) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None\
+                and cargo.weight <= self.max_load_weight:
```
danylott, line 68: use `black`

danylott: formatter I mean

danylott (changes requested):

app/main.py
```diff
@@ -1,5 +1,72 @@
...
+class BaseRobot(Cargo):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        self.name = name
         self.weight = weight
+        self.coords = coords
+        if coords is None:
+            self.coords = [0, 0]
```
wQuelS, line 18: make it in one line using ternary operator

app/main.py
```diff
@@ -1,5 +1,72 @@
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        super().__init__(name, weight)
+        self.coords = coords
+        if coords is None:
+            self.coords = [0, 0, 0]
```
wQuelS, line 44: same, and pass it right into `super`

app/main.py
```diff
@@ -1,5 +1,72 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        max_load_weight: int,
+        current_load: int,
```
wQuelS, line 59: are u sure `current_load` is `int` here ?

wQuelS (changes requested):

app/main.py
```diff
@@ -1,5 +1,68 @@
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = [0, 0, 0] if coords is None else coords
```
YuriiKindrat, line 40: u can make this check inside super method, so won't need to rewrite an attribute

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,5 +1,69 @@
...
-    def __init__(self, weight):
+    def __init__(self, weight: int) -> None:
+        self.weight = weight
+
+
+class BaseRobot(Cargo):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
```
Serhii-Leonenko, line 13: coords could be both types, so use annotation like this. The same in all other similar cases

```suggestion
class BaseRobot(Cargo):
    def __init__(self,
                 name: str,
                 weight: int,
                 coords: list[int] | None = None) -> None:
```

app/main.py
```diff
@@ -1,5 +1,69 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        max_load_weight: int,
+        current_load: Cargo,
```
Serhii-Leonenko, line 56: It's better to add a default value ` None `for current_load

Serhii-Leonenko (changes requested):

PR #339, opened 2023-02-11
Solution DonchykD

DmytryTong:


usrofgh (approved):
Fine

app/main.py
```diff
@@ -1,5 +1,67 @@
...
+
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        coords = coords if coords else [0, 0, 0]
+        super().__init__(name, weight, coords)
+
+    def go_up(self, step: int = 1) -> None:
+
```
YuriiKindrat, line 41: Remove this empty line, please, stick to the same code style everywhere

app/main.py
```diff
@@ -1,5 +1,67 @@
...
+                 current_load: None | Cargo,
+                 coords: list[int] = None) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+
+        weight = cargo.weight
+
```
YuriiKindrat, line 62: no need to create this variable + empty lines

YuriiKindrat (changes requested):

app/main.py
DmytryTong: Fixed.

app/main.py
```diff
@@ -1,5 +1,63 @@
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        coords = coords if coords else [0, 0, 0]
```
YuriiKindrat, line 37: u can write this condition inside super.init function

YuriiKindrat (changes requested):

app/main.py
DmytryTong: Fixed.

app/main.py
```diff
@@ -1,5 +1,62 @@
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords else [0, 0]
```
YuriiKindrat, line 13: this solution will work here, because it’s forbidden to have [] coords, but in other situation, if you check just if coords, but not if coords is None - then you won’t be able to specify empty array into coords at all

app/main.py
```diff
@@ -1,5 +1,62 @@
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        super().__init__(name, weight, coords if coords else [0, 0, 0])
```
YuriiKindrat, line 37: the same is for this check

YuriiKindrat (changes requested):

app/main.py
DmytryTong: Removed [].

app/main.py
DmytryTong: Removed [].

wQuelS (approved):
That is nice

PR #340, opened 2023-02-11
Solution

usrofgh:


Yevheniy-Lototskiy (approved):

app/main.py
```diff
@@ -1,5 +1,68 @@
...
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords else [0, 0]
```
danylott, line 15: this solution will work here, because it's forbidden to have `[]` coords, but in other situation, if you check just `if coords`, but not `if coords is None` - then you won't be able to specify empty array into coords at all
image

danylott (changes requested):

app/main.py
```diff
@@ -1,5 +1,68 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo,
```
wQuelS, line 56: default should be set to `none` here

wQuelS (changes requested):

app/main.py
```diff
@@ -1,5 +1,68 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: None,
```
wQuelS, line 56: It can be `Cargo` as well :))

wQuelS (changes requested):

app/main.py
usrofgh: oops...

wQuelS (approved):
Nice job

ykopatko (approved):
gj

PR #341, opened 2023-02-11
'Solution'

alexlega:


app/main.py
```diff
@@ -1,5 +1,75 @@
...
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
+        else:
+            self.coords = coords
+
+    def go_up(self, up: int = 1) -> None:
+        self.coords[2] += up
+
+    def go_down(self, up: int = 1) -> None:
```
ramsty, line 53: I think here logically positional variable have to be "down"

alexlega: yeah, thanks, its famous "ctrl + c -> ctrl + v" problem)

app/main.py
```diff
@@ -1,5 +1,75 @@
...
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        self.name = name
         self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
yura-hudzovskyi, line 19: You can do it in 1 line
`self.coords = [0, 0] if coords is None else coords`

app/main.py
```diff
@@ -1,5 +1,75 @@
...
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
+        else:
+            self.coords = coords
```
yura-hudzovskyi, line 48: same

yura-hudzovskyi (changes requested):

app/main.py
alexlega: okay, thanks for the advice)

oigod (approved):

app/main.py
```diff
@@ -1,5 +1,72 @@
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = [0, 0, 0] if coords is None else coords
```
YuriiKindrat, line 45: what if coords will be [0,0]?

YuriiKindrat (changes requested):

app/main.py
alexlega: yeah, coords will be [0, 0] in this case, I'll add condition when len(coords) != 3

wQuelS (changes requested):
Resolve conflicts

app/main.py
```diff
@@ -1,5 +1,73 @@
...
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        super().__init__(name, weight)
+        self.coords = [0, 0, 0] \
+            if coords is None or len(coords) != 3 else coords
```
wQuelS, line 46: First: try to avoid line breaks.
Second: it is bad practice to override something after `super`
Third: you can pass this as online right into the `super`

wQuelS (changes requested):

app/main.py
```diff
@@ -1,5 +1,79 @@
...
+            name: str,
+            weight: int,
+            coords: list = None,
+            max_load_weight: int = None,
+            current_load: Cargo = None
+    ) -> None:
+        super().__init__(
+            name,
+            weight,
+            [0, 0, 0] if coords is None or len(coords) != 3 else coords
```
YuriiKindrat, line 69: no need to make this check here, it is already checked in parent class

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,5 +1,79 @@
...
+            name: str,
+            weight: int,
+            coords: list = None,
+            max_load_weight: int = None,
+            current_load: Cargo = None
+    ) -> None:
+        super().__init__(
+            name,
+            weight,
+            [0, 0, 0] if coords is None else coords
```
YuriiKindrat, line 69: u didn't fix that, coords is checked in parent class, why do u need to check them here ?

YuriiKindrat (changes requested):

app/main.py
alexlega: okay, i got it, i'll fix now

YuriiKindrat (approved):

PR #345, opened 2023-02-12
Solution

yura-hudzovskyi:


amber-marichi (approved):
very nice. approved

app/main.py
```diff
@@ -1,5 +1,62 @@
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int | float,
+                 coords: list[int | float] = None) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = [0, 0, 0] if coords is None else coords
```
YuriiKindrat, line 37: what if coords will be [0,0] ?

app/main.py
```diff
@@ -1,5 +1,63 @@
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int | float,
+                 coords: list[int | float] = None) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = [0, 0, 0] if coords is None or len(coords) != 3 \
+            else coords
```
YuriiKindrat, line 38: no need to rewrite attribute here, u can just make a check for coords argument in super method

YuriiKindrat (changes requested):

PR #346, opened 2023-02-12
Solution

viktoria-rybenchuk:


app/main.py
```diff
@@ -1,5 +1,71 @@
...
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+        if self.coords is None:
```
Yulia-Mala, line 16: not a big deal but you can write it as "if not self.coords:" - it's the same

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        super().__init__(name, weight, coords=coords)
```
Yulia-Mala, line 42: not necessary  to make coords named arg here. why other args are positional  and only coords named? not a big deal as well but still

Yulia-Mala (approved):

app/main.py
danylott: Btw - not the same.
`not self.coords` will work here, because it’s forbidden to have [] coords, but in other situation, if you check just if coords, but not if coords is None - then you won’t be able to specify empty array into coords at all
>>> bool([])
False

danylott: I suggest using `if self.coords is None` in such situations

app/main.py
danylott: Agree

danylott: Should be fixed

danylott (changes requested):
Close, but fix 1 moment

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+        if self.coords is None:
+            self.coords = [0, 0]
```
wQuelS, line 17: still, this is not fixed

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
```
wQuelS, line 44: make it in oneline using ternary operator and pass it right to the `super`. No need to write this separately

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: int,
```
wQuelS, line 59: are u sure `current_load` is `int` here ?

wQuelS (changes requested):

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: None,
```
wQuelS, line 59: It can be `Cargo` or `none`

wQuelS (changes requested):

wQuelS (approved):
Good

PR #348, opened 2023-02-12
solution

amber-marichi:


app/main.py
```diff
@@ -1,5 +1,64 @@
...
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
```
ZhAlexR, line 34: I am not sure, but supper() should be called first.
But it is only suggestion.

ZhAlexR (approved):

app/main.py
```diff
@@ -1,5 +1,64 @@
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
YuriiKindrat, line 13: you can write this in one line. Don't use 3-level nested conditions. Use a single composite `if` statement. Use [pep8 rules](https://peps.python.org/pep-0008/#indentation) to make it correctly.

app/main.py
```diff
@@ -1,5 +1,64 @@
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
```
YuriiKindrat, line 35: also try to write it in one line

YuriiKindrat (changes requested):

wQuelS (approved):
GJ

PR #351, opened 2023-02-13
Solution

ghost:


app/main.py
```diff
@@ -1,5 +1,78 @@
...
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+        if self.coords is None:
+            self.coords = [0, 0]
+
+    def go_forward(self, step: int = 1) -> list:
+        self.coords[1] += step
+        return self.coords
```
MikhailLyvak, line 21: This func`s shouldn`t return anything. Only get_info returns str.

MikhailLyvak (changes requested):

app/main.py
```diff
@@ -1,5 +1,66 @@
...
+        weight: int,
+        coords: list = None,
+        max_load_weight: int = None,
+        current_load: int = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> int | bool:
```
YuriiKindrat, line 61: is this method return smth ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #352, opened 2023-02-13
Solution

ZhAlexR:


app/main.py
```diff
@@ -1,5 +1,64 @@
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
Anatolii-Poznyak, line 13: u can do it in one_line ))

Anatolii-Poznyak (approved):
GJ !

app/main.py
ZhAlexR: Agree)

app/main.py
```diff
@@ -1,5 +1,58 @@
...
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        max_load_weight: int,
+        current_load: int,
+        coords: list = None,
+    ) -> None:
+        super().__init__(name, weight, coords=coords)
```
danylott, line 49: why only 3rd argument is provided as named argument? 
Please redo this to all positional or all keyword

app/main.py
```diff
@@ -1,5 +1,58 @@
...
+        self.coords[0] -= point_x
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight)
+        self.coords = [0, 0, 0] if coords is None else coords
```
danylott, line 31: Optionally, you can think, how to just provide coords to super init, instead of overriding it after

danylott (changes requested):

app/main.py
```diff
@@ -1,5 +1,61 @@
...
+        self.coords[2] -= point_z
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        max_load_weight: int,
+        current_load: int,
```
wQuelS, line 49: are u sure `current_load` is `int` here ?

wQuelS (changes requested):

wQuelS (approved):
Good

PR #353, opened 2023-02-13
Solution

oigod:


AlessaQ (approved):

app/main.py
```diff
@@ -1,5 +1,68 @@
-class Cargo:
-    def __init__(self, weight):
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None
+                 ) -> None:
```
danylott, line 6: It's better to have arguments on each line like this: (as `black` suggests)
image

app/main.py
```diff
@@ -1,5 +1,68 @@
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None
+                 ) -> None:
+        super().__init__(name, weight, coords)
+
+        self.coords = [0, 0, 0] if coords is None else coords
```
danylott, line 35: Could you write this code please without overwriting coords after super init, but to send there correct ones

app/main.py
```diff
@@ -1,5 +1,68 @@
...
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: object,
+                 coords: list = None,
+                 ) -> None:
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+        super().__init__(name, weight, coords)
+
+    def hook_load(self, cargo: object) -> None:
```
danylott, line 56:
```suggestion
    def hook_load(self, cargo: "Cargo") -> None:
```
you can use strings, if class is lower, or even better - just move that class higher

danylott (changes requested):

wQuelS (changes requested):
please resolve conflicts first

app/main.py
wQuelS: This is still not fixed :(

app/main.py
```diff
@@ -1,5 +1,66 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: object,
```
wQuelS, line 54: everything is object in python, give more descriptive type hints

app/main.py
```diff
@@ -1,5 +1,66 @@
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        super().__init__(name, weight, coords=coords or [0, 0, 0])
```
wQuelS, line 39: you have 2 positional arguments, and only `coords` is key arg. Fix this, please

wQuelS (changes requested):

wQuelS (changes requested):
Tests are failed now :(

app/main.py
```diff
@@ -1,5 +1,67 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo,
```
wQuelS, line 55: it also can be `None` according to the task description

wQuelS (changes requested):

app/main.py
oigod: tempest double on cd , i tired of fixing xD

wQuelS (approved):
ez for Arc Warden

oigod:
> ez for Arc Warden

u are the best man. pls make review for me every time i u can <3

PR #354, opened 2023-02-13
Solution

MikhailLyvak:


ada-krav (approved):

app/main.py
```diff
@@ -1,5 +1,57 @@
...
+        self.coords[2] -= number
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        max_load_weight: int,
+        current_load: int | float,
```
YuriiKindrat, line 45: image
are u sure about int | float ?)

app/main.py
```diff
@@ -1,5 +1,57 @@
...
+        coords: list = None
+    ) -> None:
+        super().__init__(name=name, weight=weight, coords=coords or [0, 0, 0])
```
YuriiKindrat, line 48: no need to check coords here

YuriiKindrat (changes requested):

PR #355, opened 2023-02-13
Solution

Litvinenko23:


app/main.py
```diff
@@ -1,5 +1,58 @@
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords:
+            self.coords = coords
```
BohdanLazaryshyn, line 11: you can write it in one sentence

BohdanLazaryshyn: I meant if...else...

dogmoony: agree, you can do $10-13 in one line

dogmoony (changes requested):

dogmoony (approved):

app/main.py
```diff
@@ -1,5 +1,55 @@
 class Cargo:
-    def __init__(self, weight):
+    def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Serhii-Leonenko, line 7: coords could have both types of values, use annotation like this. The same in all other similar cases

```suggestion
    def __init__(self, name: str, weight: int, coords: list | None= None) -> None:
```

app/main.py
```diff
@@ -1,5 +1,55 @@
...
+    def go_down(self, value: int = 1) -> None:
+        self.coords[2] -= value
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: bool,
```
Serhii-Leonenko, line 44: bool for current load, are you sure?)

Serhii-Leonenko (changes requested):

YuriiKindrat (approved):

PR #356, opened 2023-02-13
Solution

BohdanLazaryshyn:


oksanaaam (approved):

app/main.py
```diff
@@ -1,5 +1,62 @@
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int | float,
+                 max_load_weight: int,
+                 current_load: None,
+                 coords: list = None
+                 ) -> None:
```
YuriiKindrat, line 52: fix indentation here, please
image

app/main.py
```diff
@@ -1,5 +1,62 @@
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int | float,
+                 coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords else [0, 0]
```
YuriiKindrat, line 13: this solution will work here, because it’s forbidden to have [] coords, but in other situation, if you check just if coords, but not if coords is None - then you won’t be able to specify empty array into coords at all

app/main.py
```diff
@@ -1,5 +1,62 @@
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int | float,
+                 coords: list = None) -> None:
+        super().__init__(name, weight, coords if coords else [0, 0, 0])
```
YuriiKindrat, line 36: the same here

YuriiKindrat (changes requested):

wQuelS (approved):
GJ

PR #358, opened 2023-02-13
Solution

Anatolii-Poznyak:


app/main.py
```diff
@@ -1,5 +1,58 @@
 class Cargo:
-    def __init__(self, weight):
+    def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
NazarHladaniuk, line 7: better to use 
def __init__(
            self,
            name: str,
            weight: int,
            coords: list = None
    ) -> None:
for readability

app/main.py
```diff
@@ -1,5 +1,58 @@
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
NazarHladaniuk, line 29: here too

NazarHladaniuk (changes requested):

app/main.py
Anatolii-Poznyak: it passes flake8 test, actually, so this is a matter of taste, until the mentor says otherwise

NazarHladaniuk (approved):

app/main.py
```diff
@@ -1,5 +1,58 @@
...
+    def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if isinstance(coords, list) else [0, 0]
```
wQuelS, line 10: check `if coords is None`, no need for type checking:)

app/main.py
```diff
@@ -1,5 +1,58 @@
...
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = coords if isinstance(coords, list) else [0, 0, 0]
```
wQuelS, line 31: same. And pass it right to the `super`

wQuelS (changes requested):

app/main.py
Anatolii-Poznyak: +

YuriiKindrat (approved):

PR #361, opened 2023-02-13
Solution

ada-krav:


ostapT (approved):
gj!

app/main.py
```diff
@@ -1,5 +1,76 @@
...
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
wQuelS, line 19: it can be done in one line using ternary operator

app/main.py
```diff
@@ -1,5 +1,76 @@
...
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+
+        if coords is None:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords[0:2])
+        self.coords.append(coords[2])
```
wQuelS, line 48: same. One line with ternary operator. You can pass such oneliner right into your `super`

app/main.py
```diff
@@ -1,5 +1,76 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        max_load_weight: int,
+        current_load: int,
```
wQuelS, line 63: are u sure `current_load` is `int` here ?

wQuelS (changes requested):

app/main.py
ada-krav: I`m not sure. hook_load takes Cargo, but README never tells that current_load should specifically take Cargo. But I guess it makes sense.

Serhii-Leonenko (approved):
Good job!

PR #362, opened 2023-02-13
Solution

ostapT:


ProdyRodion (approved):
i think that's good job!

app/main.py
```diff
@@ -1,5 +1,72 @@
...
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
YuriiKindrat, line 18: you can write this in one line. Don't use 3-level nested conditions. Use a single composite `if` statement. Use [pep8 rules](https://peps.python.org/pep-0008/#indentation) to make it correctly.

YuriiKindrat (changes requested):

Serhii-Leonenko (changes requested):

wQuelS (approved):
Good

PR #363, opened 2023-02-13
Solution

RomanRega:


app/main.py
```diff
@@ -1,5 +1,69 @@
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self, name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        if self.coords == [0, 0]:
```
AnnaBoichenko, line 41: maybe it`s better delete coords from super().__init__(name, weight, coords)
and write   if self.coords is None:  self.coords = [0, 0, 0]

app/main.py
```diff
@@ -1,5 +1,69 @@
...
+        weight: int,
+        max_load_weight: int,
+        current_load: int = None,
+        coords: list = [0, 0, 0],
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: int) -> None:
```
AnnaBoichenko, line 64: cargo is not int, cargo is Cargo instance!

AnnaBoichenko (changes requested):

app/main.py
RomanRega: oo, thanks

app/main.py
RomanRega: sorry but it doesnt work in my code, but i change my code in another way

AnnaBoichenko (approved):

app/main.py
```diff
@@ -1,5 +1,69 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        max_load_weight: int,
+        current_load: int = None,
```
YuriiKindrat, line 57: are u sure that current_load is int ?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,5 +1,69 @@
...
+    def __init__(
+            self, name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+        if self.coords is None:
+            self.coords = [0, 0]
```
wQuelS, line 16: can u do this in one line?

app/main.py
wQuelS, line 41: make it in one line and put it right into the `super`

app/main.py
```diff
@@ -1,5 +1,69 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        max_load_weight: int,
+        current_load: None = None,
```
wQuelS, line 57: `current_load` can be `Cargo` here with default set to `None`

wQuelS (changes requested):

wQuelS (approved):
Nice

PR #364, opened 2023-02-13
'Solution'

oksanaaam:


me1nyk (approved):
Well done

app/main.py
```diff
@@ -1,5 +1,58 @@
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None,
+            max_load_weight: int = 0,
+            current_load: None = None
```
YuriiKindrat, line 47: Are u sure that current_load should be only None ?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,5 +1,58 @@
...
+        self.coords[0] -= step
+
+    def get_info(self, step: int = 1) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight)
+        self.coords = coords or [0, 0, 0]
```
YuriiKindrat, line 31: no need to rewrite attribute here, u can just make a check for it in super method, also don't forget about is None as in Baserobot check

YuriiKindrat (changes requested):

wQuelS (changes requested):

app/main.py
```diff
@@ -1,5 +1,61 @@
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None,
+            max_load_weight: int | float = 0,
+            current_load: int | float = None
```
wQuelS, line 50: it can be `Cargo` or `None`. You can read about it in task description

wQuelS (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #365, opened 2023-02-13
Solution

me1nyk:


app/main.py
```diff
@@ -1,5 +1,62 @@
...
+
+    def go_left(self, point_x: int = 1) -> None:
+        self.coords[0] -= point_x
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
destr34, line 32: split this line, like in DeliveryDrone class

app/main.py
```diff
@@ -1,5 +1,62 @@
 class Cargo:
-    def __init__(self, weight):
+    def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
destr34, line 7: the same, split line

destr34 (changes requested):

app/main.py
me1nyk: I dont think its necessary, because __init__ for DeliveryDrone  class is too long, when for BaseRobot and FlyingRobot just 3, but okay)

app/main.py
```diff
@@ -1,5 +1,72 @@
...
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
YuriiKindrat, line 18: you can write this in one line. Don't use 3-level nested conditions. Use a single composite `if` statement. Use [pep8 rules](https://peps.python.org/pep-0008/#indentation) to make it correctly.

app/main.py
```diff
@@ -1,5 +1,72 @@
...
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
```
YuriiKindrat, line 45: u can make this check inside the super method

app/main.py
```diff
@@ -1,5 +1,72 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        max_load_weight: int,
+        current_load: int,
```
YuriiKindrat, line 60: are u sure, that current_load will be int?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,5 +1,67 @@
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
```
Serhii-Leonenko, line 11: coords could be both types, so use annotation like this. The same in all other similar cases

```suggestion
            coords: list | None = None
```

app/main.py
```diff
@@ -1,5 +1,67 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        max_load_weight: int,
+        current_load: Cargo,
```
Serhii-Leonenko, line 55: It's better to add a default value None for current_load. This will allow us Drone instances without redundant arguments

Serhii-Leonenko (changes requested):

app/main.py
me1nyk: when i am using annotation like this, tests are failed :(
image

app/main.py
me1nyk: Thanks!

wQuelS (approved):
Nice

PR #367, opened 2023-02-13
'Solution'

NazarHladaniuk:


app/main.py
```diff
@@ -1,5 +1,69 @@
...
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords else [0, 0]
```
oksana-feshchenko, line 15: You should check if coords is not None, because otherwise, it will be impossible to put empty list for coords

app/main.py
```diff
@@ -1,5 +1,69 @@
...
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = coords if coords else [0, 0, 0]
+
```
oksana-feshchenko, line 42: first of all it will be better to move this check in super().__init__( here!!!)
and check with None the same as before

app/main.py
```diff
@@ -1,5 +1,69 @@
...
+            coords: list = None,
+            current_load: int = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None:
+            if cargo.weight <= self.max_load_weight:
```
oksana-feshchenko, line 65: you could do this check in one line

oksana-feshchenko (changes requested):

oksana-feshchenko (approved):

YuriiKindrat (changes requested):

Nattalli (approved):

PR #368, opened 2023-02-13
Solution

TeliksPG:


oksanaaam (approved):

app/main.py
```diff
@@ -1,5 +1,74 @@
...
+        self.coords[1] -= y_coordinate
+        return self.coords
+
+    def go_right(self, x_coordinate: int = 1) -> list:
+        self.coords[0] += x_coordinate
+        return self.coords
+
+    def go_left(self, x_coordinate: int = 1) -> list:
+        self.coords[0] -= x_coordinate
+        return self.coords
```
YuriiKindrat, line 31: no need to return coords here

app/main.py
```diff
@@ -1,5 +1,74 @@
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        super().__init__(name, weight)
+        self.coords = [0, 0, 0] if coords is None else coords
```
YuriiKindrat, line 45: u can make this check inside super func, no need to rewrite attribute

app/main.py
```diff
@@ -1,5 +1,74 @@
...
+
+    def go_up(self, z_coordinate: int = 1) -> list:
+        self.coords[2] += z_coordinate
+        return self.coords
+
+    def go_down(self, z_coordinate: int = 1) -> list:
+        self.coords[2] -= z_coordinate
+        return self.coords
```
YuriiKindrat, line 53: the same is for this returns

app/main.py
```diff
@@ -1,5 +1,74 @@
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        max_load_weight: int,
+        current_load: int | None,
```
YuriiKindrat, line 62: are u sure that current_load is int ?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,5 +1,70 @@
...
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        super().__init__(
+            name,
+            weight,
+            coords=[0, 0, 0] if coords is None else coords)
```
Serhii-Leonenko, line 43: fix indentation here

app/main.py
```diff
@@ -1,5 +1,70 @@
...
+        self.coords[2] -= z_coordinate
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        max_load_weight: int,
+        current_load: None,
```
Serhii-Leonenko, line 58: I don't agree with the annotation for current_load, also use the default value for it

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1,5 +1,71 @@
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
```
Serhii-Leonenko, line 11: Sorry, I didn't say you before. But coords could be both types, for that purpose use annotation like this. The same in other similar cases

```suggestion
            coords: list | None = None
```

Serhii-Leonenko (changes requested):

wQuelS (approved):
Nice

PR #370, opened 2023-02-13
'Solution'

ykopatko:


app/main.py
```diff
@@ -1,5 +1,74 @@
...
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int | float,
+            coords: list = None
+    ) -> None:
+        if coords is None:
+            coords = [0, 0]
```
oksana-feshchenko, line 14: You could do it in one line using ternary operators

app/main.py
```diff
@@ -1,5 +1,74 @@
...
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int | float,
+            coords: list = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
```
oksana-feshchenko, line 44: make this check inside super in one line

app/main.py
oksana-feshchenko: example attr if attr is not None else some_value

oksana-feshchenko (changes requested):

oksana-feshchenko (approved):

wQuelS (changes requested):
Resolve all the conflicts

ykopatko:
> 

I do not understand what these conflicts mean

ykopatko:
> Resolve all the conflicts

I have tried something just recently. Hope that will work

app/main.py
```diff
@@ -1,5 +1,72 @@
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int | float,
+            coords: list = None
```
Serhii-Leonenko, line 11: The value could be both types, so use an annotation like this. The same in other similar cases


```suggestion
    def __init__(
            self,
            name: str,
            weight: int | float,
            coords: list | None= None
```

Serhii-Leonenko (changes requested):

YuriiKindrat (approved):

PR #371, opened 2023-02-13
Solution

Bohdan-Kovalenko1992:


app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if not coords:
+            coords = [0, 0]
```
oksana-feshchenko, line 11: You should check if coords is not None, because otherwise, it will be impossible to put empty list for coords
and make it in one line using ternary operators
example  attr if attr is not None else some_value

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        if not coords:
+            self.coords = [0, 0, 0]
```
oksana-feshchenko, line 34: first of all it will be better to move this check in super().init( here!!!)
and check with None the same as before

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int | Cargo | None,
+                 coords: list = None
+                 ) -> None:
+        super().__init__(name, weight, coords)
+        if not coords:
+            self.coords = [0, 0, 0]
```
oksana-feshchenko, line 53: the same here and move this check inside super

oksana-feshchenko (changes requested):

app/main.py
Bohdan-Kovalenko1992: its actually i what i want. If list is empty or none its come [0, 0].
Ternary bad for reading. I dont want use it here.

app/main.py
Bohdan-Kovalenko1992: are it really better?

app/main.py
Bohdan-Kovalenko1992: agree

app/main.py
oksana-feshchenko: But about check with None it was from mentor, so I'll leave it up to you

oksana-feshchenko (approved):

wQuelS (approved):
GJ

PR #372, opened 2023-02-13
Solution

tytomyr:


app/main.py
```diff
@@ -1,5 +1,63 @@
...
-# write your code here
+
+class BaseRobot:
+
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0]
+        self.name = name
+        self.weight = weight
+        self.coords = coords
```
ZabFTFT, line 16: You can write check condition and assignment of coords in 1 line, but its just suggestion.

app/main.py
```diff
@@ -1,5 +1,63 @@
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        self.coords = coords
+        super().__init__(name, weight, coords)
```
ZabFTFT, line 39: Same here

app/main.py
```diff
@@ -1,5 +1,63 @@
...
+    def go_up(self, step: int = 1) -> None:
+        self.coords[2] += step
+
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self, name: str, weight: int, max_load_weight: int,
+                 coords: list = None, current_load: int = None) -> None:
```
ZabFTFT, line 50: current load also can be Cargo

app/main.py
```diff
@@ -1,5 +1,63 @@
...
+        self.coords[0] += step
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
```
ZabFTFT, line 34: Make blank line after class keyword.Be consistent with your style. I got reject by Danya cause of it =))

ZabFTFT (approved):
Well done! Code is nice, there are some suggestions that mentors don't like to accept, so its up to you to fix it now or got reject later =)

app/main.py
wQuelS: Agree. Write it in one line, please

app/main.py
wQuelS: Yep.You left a blank line previously, so you'd better keep doing so

app/main.py
wQuelS: Agree, you can write it in one line and pass directly to `super`

app/main.py
wQuelS: yep

app/main.py
```diff
@@ -1,5 +1,63 @@
...
+
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self, name: str, weight: int, max_load_weight: int,
+                 coords: list = None, current_load: int = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
```
wQuelS, line 52: again, one line and put it directly to `super`

wQuelS (changes requested):

app/main.py
```diff
@@ -1,5 +1,67 @@
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        coords = [0, 0, 0] if coords is None else coords
```
wQuelS, line 41: you pass this right into the super

app/main.py
```diff
@@ -1,5 +1,67 @@
...
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: None | Cargo,
```
wQuelS, line 56: it can be `Cargo` with default value set to `None`

wQuelS (changes requested):

YuriiKindrat (approved):

PR #375, opened 2023-02-13
Solutionis

skmiwa1:


app/main.py
```diff
@@ -1,5 +1,62 @@
...
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int = None,
+                 coords: list = None) -> None:
+        super().__init__(name,
+                         weight,
+                         coords if coords is not None else [0, 0, 0])
```
manjustice, line 53: There is no need to check if coords is None because u check it in FlyingRobot class

manjustice (changes requested):

manjustice (approved):

app/main.py
```diff
@@ -1,5 +1,62 @@
...
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int = None,
```
YuriiKindrat, line 49: are u sure about type annotation for current_load ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #378, opened 2023-02-13
Solution

destr34:


app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None) -> None:
```
myroslavatrschk, line 11: Remove this empty line at line 7 and fix the indentation, please. Example:
```
def __init__(
            self,
            name: str,
            weight: int,
            coords: list = None
    ) -> None:
```

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self,
+                 name: str,
+                 weight: [int, float],
+                 coords: list = None) -> None:
```
myroslavatrschk, line 37: Same.

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(self,
+                 name: str,
+                 weight: [int, float],
+                 max_load_weight: int,
+                 current_load: None,
+                 coords: list = None) -> None:
```
myroslavatrschk, line 55: Same.

myroslavatrschk (changes requested):

myroslavatrschk (approved):

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(self,
+                 name: str,
+                 weight: [int, float],
+                 max_load_weight: int,
+                 current_load: None,
```
wQuelS, line 54: 'current_load` can be `Cargo` as well

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+                 coords: list = None) -> None:
+        super().__init__(name, weight)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+        self.coords = [0, 0, 0] if coords is None else coords
```
wQuelS, line 59: Do you need it? 
Also, it would be better to put it right into `super()`, no need to redefine smth after it

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self,
+                 name: str,
+                 weight: [int, float],
+                 coords: list = None) -> None:
+        super().__init__(name, weight)
+        self.coords = [0, 0, 0] if coords is None else coords
```
wQuelS, line 39: why not right into the `super` ?

wQuelS (changes requested):

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(self,
+                 name: str,
+                 weight: [int, float],
+                 max_load_weight: int,
+                 current_load: Cargo,
```
wQuelS, line 53: and default value should be `None` :)

wQuelS (changes requested):

app/main.py
destr34: :)

wQuelS (approved):
Good

PR #380, opened 2023-02-14
Add solution

ZaikaDaria:


app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int,
+                 coords: list = None
+                 ) -> None:
+        super().__init__(name, weight, [0, 0, 0] if coords is None else coords)
```
ArsenPidhoretskyi, line 52: [0, 0, 0] if coords is None else coords Redundant here it will be done it FlyingRobot

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
+            self.current_load = cargo
+
+    def unhook_load(self) -> object:
```
ArsenPidhoretskyi, line 60: return type is not object)

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
...
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int,
```
katevasilenko, line 48:
```suggestion
                 current_load: Cargo,
```

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str,
+                 weight: int,
+                 coords: list = None
+                 ) -> None:
```
katevasilenko, line 32:
```suggestion
    def __init__(
        self, 
        name: str,
        weight: int,
        coords: list = None
    ) -> None:
```

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
...
+        super().__init__(name, weight,
+                         [0, 0, 0] if coords is None else coords)
```
katevasilenko, line 34: `coords if coords else [0, 0, 0]` and I suppose in this way it match flake8 line length rule

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int,
+                 coords: list = None
+                 ) -> None:
```
katevasilenko, line 50: Please provide same code style as I suggested you on 32 line

katevasilenko (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #381, opened 2023-02-16
Solution

Esxoyne:


app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+        if not coords:
+            self.coords = [0, 0]
```
ArsenPidhoretskyi, line 17: Write inline this)
use `or` or inline if

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #385, opened 2023-02-20
'Solution'

iishchenko:


app/main.py
```diff
@@ -2,4 +2,71 @@ class Cargo:
...
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
+
+    def go_forward(self, step: int = 1) -> None:
+        self.coords[1] += step
```
Grayder0152, line 21: Create class variable as axis `x=0, y=1`

app/main.py
```diff
@@ -2,4 +2,71 @@ class Cargo:
...
+            name: str,
+            weight: int,
+            coords: list[int, int, int] | None = None
+    ) -> None:
+        super(FlyingRobot, self).__init__(name, weight, coords)
+        if len(self.coords) < 3:
+            self.coords.append(0)
+
+    def go_up(self, step: int = 1) -> None:
+        self.coords[2] += step
```
Grayder0152, line 48: Create class variable `z=2` to avoid unclearing logic

app/main.py
```diff
@@ -2,4 +2,71 @@ class Cargo:
...
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo | None,
+            coords: list[int, int, int] | None = None,
+    ) -> None:
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+        super(DeliveryDrone, self).__init__(name, weight, coords)
```
Grayder0152, line 65: In Python 3 to run parant`s method we can use `super()` instead `super(DeliveryDrone, self)`. Python is smart enough to define the method automatically

Grayder0152 (changes requested):

iishchenko:
I changed In Python 3 to run the parents method we can use super() instead of super (DeliveryDrone, self)`. Python is smart enough to define the method automatically.
Vasyl Smutok cancelled Create class variable as axis x=0, y=1 and Create class variable z=2 to avoid unclear logic

Grayder0152 (approved):

PR #387, opened 2023-02-22
Solution

VladyslavKaminskyi:


app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] | None = None
+                 ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords else [0, 0]
+
+    def go_forward(self, step: int = 1) -> None:
```
Grayder0152, line 16: INstead of using `1`, create constant and use it (and in another places too)

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+        self.coords[1] += step
+
+    def go_back(self, step: int = 1) -> None:
+        self.coords[1] -= step
+
+    def go_right(self, step: int = 1) -> None:
+        self.coords[0] += step
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
```
Grayder0152, line 26: Instead of using index `0`,`1` or `2`  as `x, y, z` axis, create constances for them `X_AXIS, Y_AXIS, Z_AXIS`

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 max_load_weight: int,
+                 current_load: Cargo | None,
+                 name: str,
+                 weight: int,
+                 coords: list[int] | None = None,
+                 ) -> None:
+        super().__init__(name, weight, coords=coords if coords else [0, 0, 0])
```
Grayder0152, line 55:
```suggestion
        super().__init__(name, weight, coords=coords)
```
Parent class will make all checks

Grayder0152 (changes requested):

PR #401, opened 2023-03-16
'Solution'

HalynaPetrova:


app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list | None = None) -> None:
```
vasylhnatiuk, line 11: use `Optional[list]` if argument can be `None` here and other places https://stackoverflow.com/questions/51710037/how-should-i-use-the-optional-type-hint

vasylhnatiuk (changes requested):

Vasyl-Smutok (changes requested):

Vasyl-Smutok (approved):
Well Done 👍🏻

PR #404, opened 2023-03-21
Solution_1

nickkozlov90:


app/main.py
```diff
@@ -1,5 +1,67 @@
...
+
 class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list[int] = None) \
+            -> None:
```
Bohdan-Salamakha, line 11:
```
def __init__(
        self,
        name: str,
        weight: int,
        coords: Optional[list[int]] = None
) -> None:
```

app/main.py
```diff
@@ -1,5 +1,67 @@
...
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list[int] = None) \
+            -> None:
```
Bohdan-Salamakha, line 36: format it like in other comment

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #408, opened 2023-03-25
Solution

goldenuni:


app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
```
Bohdan-Salamakha, line 10: use typehint like: Optional[YOUR_VALUE_TYPE] = None

app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
```

app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 coords: list[int] = None,
+                 current_load: Cargo = None) -> None:
```

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #426, opened 2023-04-14
solution

alim2709:


app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+        if self.coords is None:
+            self.coords = [0, 0]
```
ArsenPidhoretskyi, line 17: Write this inline please

app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        if self.coords is None or self.coords == [0, 0]:
+            self.coords = [0, 0, 0]
```
ArsenPidhoretskyi, line 44: Do before init changing coords:
coords = (and here inline if or use `or`)

app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
+
+    def go_up(self, step: int = 1) -> None:
+        if len(self.coords) == 2:
+            self.coords.append(step)
```
ArsenPidhoretskyi, line 48: It is redundant code please remove it

ArsenPidhoretskyi: The same in go_down
We know that this robot must work when len(coords) is equal to 3! We no need to check this in this method

ArsenPidhoretskyi (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #432, opened 2023-04-22
Soluttion

denysbelei9:


app/main.py
```diff
@@ -1,5 +1,58 @@
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Bohdan-Salamakha, line 2: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contain string elements has typehint list[str]

use typehint like: Optional[YOUR_VALUE_TYPE] = None

app/main.py
```diff
@@ -1,5 +1,58 @@
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Bohdan-Salamakha, line 26: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contain string elements has typehint list[str]

use typehint like: Optional[YOUR_VALUE_TYPE] = None

app/main.py
```diff
@@ -1,5 +1,58 @@
...
+        self.coords[2] += step
+
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self, name: str, weight: int,
+            coords: list = None, max_load_weight: int = 0,
```
Bohdan-Salamakha, line 41: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contain string elements has typehint list[str]

use typehint like: Optional[YOUR_VALUE_TYPE] = None

app/main.py
```diff
@@ -1,5 +1,58 @@
...
+            current_load: int = None
```
Bohdan-Salamakha, line 42: use typehint like: Optional[YOUR_VALUE_TYPE] = None

app/main.py
```diff
@@ -1,5 +1,58 @@
...
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: list) -> None:
```
Bohdan-Salamakha, line 48: cargo isn't a list

Bohdan-Salamakha (changes requested):
v1

app/main.py
denysbelei9: Done

PR #433, opened 2023-04-22
Develop

denysbelei9:


app/main.py
```diff
@@ -1,5 +1,64 @@
+class BaseRobot:
+    def __init__(
+            self, name: str,
+            weight: int, coords: list[int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            coords = [0, 0]
+        self.coords = coords
```
ArsenPidhoretskyi, line 10: Write this inline

app/main.py
```diff
@@ -1,5 +1,64 @@
+class BaseRobot:
+    def __init__(
+            self, name: str,
+            weight: int, coords: list[int] = None
```
ArsenPidhoretskyi, line 4: It accepts Optional[list[int]] because None as default argument provided. Fix in all places

ArsenPidhoretskyi (changes requested):

PR #435, opened 2023-04-23
Solution

romanfedorchuk:


app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Bohdan-Salamakha, line 7: use typehint like: Optional[YOUR_VALUE_TYPE] = None

add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contain string elements has typehint list[str]

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Bohdan-Salamakha, line 29: use typehint like: Optional[YOUR_VALUE_TYPE] = None

add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contain string elements has typehint list[str]

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            coords: list = None,
```
Bohdan-Salamakha, line 48: use typehint like: Optional[YOUR_VALUE_TYPE] = None

add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contain string elements has typehint list[str]

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
...
+            current_load: Cargo = None) -> None:
```
Bohdan-Salamakha, line 49: use typehint like: Optional[YOUR_VALUE_TYPE] = None

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
...
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, obj: Cargo) -> None:
+        if self.max_load_weight >= obj.weight and self.current_load is None:
+            self.current_load = obj
+
+    def unhook_load(self) -> None:
+        if self.current_load is not None:
+            self.current_load = None
```
Bohdan-Salamakha, line 60:
```suggestion
        self.current_load = None
```

Bohdan-Salamakha (changes requested):
v1

app/main.py
```diff
@@ -1,5 +1,72 @@
...
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: Optional[list[int]] = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = coords
+        if self.coords is None:
+            self.coords = [0, 0, 0]
```
ArsenPidhoretskyi, line 46:
```suggestion
        if coords is None:
            coords = [0, 0, 0]
        super().__init__(name, weight, coords)
```

ArsenPidhoretskyi: For now it is strange because you three times can change coords

app/main.py
```diff
@@ -1,5 +1,72 @@
...
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            coords: Optional[list[int]] = None,
+            current_load: Optional[Cargo] = None) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, obj: Cargo) -> None:
```
ArsenPidhoretskyi, line 67: obj -> cargo
rename

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi: @romanfedorchuk not fixed

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Good job!
Але 
box - коробка
cargo - груз
Тому ім'я не дуже ок

PR #437, opened 2023-04-24
Develop

sbilikepy:


sbilikepy:
Hi, im not sure about correct type annotation here
` list[int]` or `list[int, int]` / `list[int, int, int]`

sbilikepy:
Or should i use` coords: Union[List[int, int], List[int, int, int]]` ?

Grayder0152 (approved):

PR #440, opened 2023-04-28
Solution

Dimdimior:


app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self, name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int | None,
+                 coords: list = None) -> None:
```
Bohdan-Salamakha, line 49: use typehint like: Optional[YOUR_VALUE_TYPE] = None

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Bohdan-Salamakha, line 7: use typehint like: Optional[YOUR_VALUE_TYPE] = None

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+
+    def go_right(self, step: int = 1) -> None:
+        self.coords[0] += step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self. weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Bohdan-Salamakha, line 32: use typehint like: Optional[YOUR_VALUE_TYPE] = None

Bohdan-Salamakha (changes requested):
v1

app/main.py
Dimdimior: Done, but why is that essential?

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #442, opened 2023-04-29
Solution

jinjo0222988:


app/main.py
```diff
@@ -2,4 +2,88 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] = None) -> None:
```

app/main.py
```diff
@@ -2,4 +2,88 @@ class Cargo:
...
+        self.weight = weight
+        self.coords = coords
+        if self.coords is None:
+            self.coords = [0, 0]
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+    def go_forward(self, step: int = 1) -> None:
+        if len(self.coords) < 3:
```
vasylhnatiuk, line 18: You have much useless code, write only` self. coords[1] += step ` in this function

vasylhnatiuk: and change all of the functions

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):
Well done!

PR #447, opened 2023-05-04
drone1

melnikalex1977:


app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot(Cargo):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] = None) -> None:
```

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+                 coords: list = None, current_load: tuple = None) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
+            self.current_load = cargo
+
+    def unhook_load(self) -> tuple:
```
vasylhnatiuk, line 57: it doesn't return tuple

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot(Cargo):
+    def __init__(self, name: str, weight: int,
+                 coords: list[int] = None) -> None:
```
Grayder0152, line 8:
```suggestion
                 coords: Optional[list[int]] = None) -> None:
```

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+        return self.coords
+
+    def go_down(self, step: int = 1) -> int:
+        self.coords[2] -= step
+        return self.coords
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self, name: str, weight: int, max_load_weight: int,
+                 coords: list[int] = None, current_load: tuple = None) -> None:
```
Grayder0152, line 50: the same as above

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+        self.coords[0] -= step
+        return self.coords
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int,
+                 coords: list[int] = None) -> None:
```
Grayder0152, line 35: the same as above

Grayder0152 (approved):

PR #449, opened 2023-05-05
Solution of task "Delivery drone"

laskarj:


app/main.py
```diff
@@ -2,4 +2,78 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: [Cargo, None],
```
ArsenPidhoretskyi, line 62:
```suggestion
            current_load: Optional[Cargo],
```

app/main.py
```diff
@@ -2,4 +2,78 @@ class Cargo:
...
+            coords: list = None,
```
ArsenPidhoretskyi, line 63:
```suggestion
            coords: Optional[list[int]] = None,
```

app/main.py
```diff
@@ -2,4 +2,78 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot:
+
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
```
ArsenPidhoretskyi, line 12:
```suggestion
            coords: Optional[list[int]] = None
```
It is optional because it accepts None as default

ArsenPidhoretskyi (changes requested):
Fix type annotations

Vasyl-Smutok (approved):
Good Job 👍

PR #455, opened 2023-05-12
Solution

OksanaAdamchuk:


app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+        if coords is None:
+            self.coords = [0, 0]
```
ArsenPidhoretskyi, line 12: You can write this inline

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            self.coords = [0, 0, 0]
+            super().__init__(name=name, weight=weight, coords=self.coords)
+        else:
+            super().__init__(name=name, weight=weight, coords=coords)
```
ArsenPidhoretskyi, line 36: Duplicate of code.
Please simplify it.
If coords is None just do coords = [0, 0, 0]
And than super will be the same

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -2,4 +2,54 @@ class Cargo:
...
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self, name: str, weight: int,
+            max_load_weight: int, current_load: int,
+            coords: list = None) -> None:
+        super().__init__(name, weight, coords)
```
vasylhnatiuk, line 46:
```suggestion
        super().__init__(name=name, weight=weight, coords=coords)
```

app/main.py
```diff
@@ -2,4 +2,54 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] = None) -> None:
```

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
```
ArsenPidhoretskyi, line 9: It accepts Optional[list[int]] because None as default

ArsenPidhoretskyi: Fix in other places

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #469, opened 2023-05-24
Drones... Passes those tests at least

thorium-zero:


app/main.py
```diff
@@ -1,5 +1,109 @@
...
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.current_load = current_load
+        self.max_load_weight = max_load_weight
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is not None:
+            return
+        if cargo is None or cargo.weight > self.max_load_weight:
+            return
```
YuriiKindrat, line 103: no need to use if twice here, u can multiply them by "and", also no need to use blank returns here, u can use backward condition here

app/main.py
```diff
@@ -1,5 +1,109 @@
...
+        self.current_load = cargo
+
+    def unhook_load(self) -> None:
+        if self.current_load is None:
+            return
```
YuriiKindrat, line 108: same here

YuriiKindrat (changes requested):

app/main.py
thorium-zero: yes, I could, but those are guard clauses, they're suppose to throw exceptions or... simply return... Also, it is actually better to split those "if" statements since I check for pretty different types of issues.

app/main.py
```diff
@@ -1,5 +1,107 @@
...
+        self._coords[Axis.Zc] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        max_load_weight: int,
+        current_load: int,
```
YuriiKindrat, line 92: it seems to me, that current loas is not int )

app/main.py
```diff
@@ -1,5 +1,107 @@
...
+class BaseRobot(ABC):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: list[int] = None
+    ) -> None:
+        self._name: str = name
+        self._weight: int = weight
+        self._coords: list[int] = coords if coords is not None else [0, 0]
```
YuriiKindrat, line 34: also u can simplify this like this `coords or [0,0]` this will also check for blank list

YuriiKindrat (changes requested):

app/main.py
thorium-zero: damn right it isn't :D

PR #478, opened 2023-05-30
Delivery drones

FadieievaA:


app/main.py
```diff
@@ -2,4 +2,71 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: float,
+            coords: list = None
```
EdAlekseiev, line 11:
```suggestion
            coords: list[int] | None = None
```

EdAlekseiev: If your variable can be None, so you must use the next type hints `Optional[VALUE_TYPE]` or `YOUR_VALUE_TYPE | None`.
Also, add additional type hint for list inner objects like `list[VALUE_TYPE]`

app/main.py
```diff
@@ -2,4 +2,71 @@ class Cargo:
...
+                f"Weight: {self.weight}")
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(
+            self,
+            name: str,
+            weight: float,
+            coords: list = None
```
EdAlekseiev, line 40: The same issue as above

app/main.py
```diff
@@ -2,4 +2,71 @@ class Cargo:
...
+
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(
+            self,
+            name: str,
+            weight: float,
+            max_load_weight: float,
+            current_load: Cargo,
+            coords: list = None
```
EdAlekseiev, line 60: The same issue as above

YukihiroSM (approved):
Nice job

PR #479, opened 2023-05-31
finish Delivery drone  lab

XOctOpus1:


app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
+        else:
+            self.coords = coords
```
ArsenPidhoretskyi, line 37: Write this inline

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -2,4 +2,56 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] = None) -> None:
```

vasylhnatiuk: add more accurate annotation here and all of the code

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #485, opened 2023-06-02
'Solution'

struhanets:


app/main.py
```diff
@@ -1,5 +1,75 @@
+from __future__ import annotations
+
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None
```
vsmutok, line 8: use typehint like: Optional[YOUR_VALUE_TYPE] = None

app/main.py
```diff
@@ -1,5 +1,75 @@
...
+                 ) -> None:
+        self.name = name
+        self.weight = weight
+
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
vsmutok, line 16: try to write it in one line 
use `or`

vsmutok (changes requested):

app/main.py
```diff
@@ -1,5 +1,74 @@
+from __future__ import annotations
+
+from typing import Optional
+
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: Optional[list[str]] = None
```
ArsenPidhoretskyi, line 10: It accepts Optional[list[int]] not str
Fix in all places

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Well done

PR #487, opened 2023-06-02
Solution

Andriy-Sydorenko:


app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
...
+
+class BaseRobot:
+
+    def __init__(self, name: str, weight: 34, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
YuriiKindrat, line 14: try to simplify this using `or` so u can write this in one line

app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
...
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
+        else:
+            self.coords = coords
```
YuriiKindrat, line 39: same, also u can make this inside super

app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int,
```
YuriiKindrat, line 54: are u sure about int ?

YuriiKindrat (changes requested):

app/main.py
Andriy-Sydorenko: > are u sure about int ?

you mean that I need to change max_load_weight and current_load type annotations to float?

LiudmylaKulinchenko (approved):
Well done 😎

PR #488, opened 2023-06-02
Solution

Tarasidze:


app/main.py
```diff
@@ -1,5 +1,65 @@
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: None = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords:
+            self.coords = coords
+        else:
+            self.coords = [0, 0]
```
YuriiKindrat, line 16: try to rewrite this in one line
hint: u can use `or`

app/main.py
```diff
@@ -1,5 +1,65 @@
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: None = None) -> None:
+        super().__init__(name=name, weight=weight, coords=coords)
+        if coords:
+            self.coords = coords
+        else:
+            self.coords = [0, 0, 0]
```
YuriiKindrat, line 40: same here, also u can make this inside super method

app/main.py
```diff
@@ -1,5 +1,65 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: None = None,
+                 max_load_weight: int = None,
+                 current_load: int = None) -> None:
```
YuriiKindrat, line 55: are u sure about type annotation here?

YuriiKindrat (changes requested):

app/main.py
Tarasidze: thanks for the advice, that looks really better and more readable

app/main.py
```diff
@@ -1,5 +1,59 @@
...
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name=name, weight=weight, coords=coords)
+        self.coords = coords or [0, 0, 0]
```
YuriiKindrat, line 34: no need to rewrite attribute , u can make check inside super method)

YuriiKindrat (approved):

PR #489, opened 2023-06-03
solved

oleksiikolii:


app/main.py
```diff
@@ -1,5 +1,66 @@
+from __future__ import annotations
```
LiudmylaKulinchenko, line 1: Do you need this import?

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🎈☀

PR #492, opened 2023-06-04
Solution

DmytroTsehelnyk:


app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: list[int] = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = coords if coords is not None else [0, 0, 0]
```
YuriiKindrat, line 41: same, also u can make this inside super

YuriiKindrat (changes requested):

PR #495, opened 2023-06-04
'Solution'

yakhlamova:


app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+        if coords is None:
+            self.coords = [0, 0]
```
YuriiKindrat, line 12: try to rewrite this in one line
hint: u can use `or`

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
```
YuriiKindrat, line 34: same, also u can make this inside super method

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None,
+                 max_load_weight: int = None,
+                 current_load: int = None) -> None:
```
YuriiKindrat, line 49: are u sure about type annotation for current_load?

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, obj: Cargo) -> None:
+        if self.current_load is None and obj.weight <= self.max_load_weight:
+            self.current_load = obj
```
YuriiKindrat, line 56: why don't u name it cargo instead of obj ?

YuriiKindrat (changes requested):

app/main.py
yakhlamova: "hook_load method taking Cargo object" so I just named it like obj for better understanding. I`ll change.

LiudmylaKulinchenko (approved):
Good job 🔥

PR #502, opened 2023-06-05
Solution

Gvilion:


app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0]
```
YuriiKindrat, line 9: try to simplify this using `or` so u can write this in one line

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
```
YuriiKindrat, line 34: same, also u can write this inside super)

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        coords = coords or [0, 0]
```
YuriiKindrat, line 8: could u make this near coords attr ?)

YuriiKindrat (changes requested):

PR #503, opened 2023-06-05
Solution

Kirontiko:


app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if not coords:
+            coords = [0, 0]
```
YuriiKindrat, line 9: try to simplify this using `or` so u can make this in one line

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if not coords:
+            coords = [0, 0, 0]
```
YuriiKindrat, line 34: same is here

app/main.py
Kirontiko: > try to simplify this using `or` so u can make this in one line

Oh, didn't know about this method, thx

app/main.py
```diff
@@ -1,5 +1,65 @@
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int | float,
+                 coords: list = None) -> None:
+
+        coords = coords or [0, 0, 0]
+        super().__init__(name, weight, coords)
```
YuriiKindrat, line 39: could u make this inside super method ?))

app/main.py
```diff
@@ -1,5 +1,65 @@
...
+    def go_down(self, z_axis: int = 1) -> None:
+        self.coords[2] -= z_axis
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int | float,
+                 max_load_weight: int | float,
+                 current_load: int | float,
```
YuriiKindrat, line 53: are u sure about int \ float, what do u write in this attr inside method hook_load ?)

YuriiKindrat (changes requested):

app/main.py
Kirontiko: I did, but then I thought that it will look ugly. I will fix it

app/main.py
Kirontiko: > are u sure about int \ float, what do u write in this attr inside method hook_load ?)

oh, sorry

YuriiKindrat (approved):

PR #504, opened 2023-06-05
Solution

erikagayan:


app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+
+        if coords is None:
+            coords = [0, 0]
+        self.coords = coords
```
YuriiKindrat, line 13: try to rewrite this in one line
hint: u can use `or`

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+
+        if coords is None:
+            coords = [0, 0, 0]
+        self.coords = coords
```
YuriiKindrat, line 37: also there is no need to rewrite attribute here, u can do this inside super

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #508, opened 2023-06-05
'Solution'

MarinaKhrunina:


app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot(Cargo):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(weight)
+        self.name = name
+        self.coords = [0, 0] if coords is None else coords
```
vsmutok, line 10: I think better use `or` in this case

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
...
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = [0, 0, 0] if coords is None else coords
```
vsmutok, line 31: I think better use `or` in this case

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
vsmutok, line 29:
```suggestion
    def __init__(self, name: str, weight: int, coords: list[???] = None) -> None:
```

vsmutok (changes requested):

app/main.py
MarinaKhrunina: I don't exactly understand where? PyCharm insists on else(

vsmutok: `self.coords = coords or [0, 0]`

vsmutok (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #512, opened 2023-06-07
Solution delivery drone

MarkShapka:


app/main.py
```diff
@@ -1,5 +1,62 @@
...
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self, name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 coords: list[Coord] = None,
+                 current_load: Cargo = None) -> None:
```
Bohdan-Salamakha, line 52: use typehint like: Optional[YOUR_VALUE_TYPE] = None

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #522, opened 2023-06-18
Solution

NazarLytvyniuk:


app/main.py
```diff
@@ -2,4 +2,76 @@ class Cargo:
...
+            weight: int,
+            max_load_weight: int,
+            current_load: int,
+            coords: list[int] = None
+    ) -> None:
+
+        super().__init__(
+            name=name,
+            weight=weight,
+            coords=(coords if coords else [0, 0, 0])
```
Dimosphen1, line 67: There is no need to implement this logic `(coords if coords else [0, 0, 0])` twice, assigning [0, 0, 0] is implemented in the `FlyingRobot` class. You can just pass `coords=coords` here.

app/main.py
```diff
@@ -2,4 +2,76 @@ class Cargo:
...
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+
+        super().__init__(
+            name=name,
+            weight=weight,
+            coords=(coords if coords else [0, 0, 0])
```
Dimosphen1, line 44: It would be better to stick to one way of using the `if ... else` condition, either with `()` or without as on line 15. The more common use is without `()`.

Dimosphen1 (approved):
Everything looks fine, two minor comments left.

app/main.py
NazarLytvyniuk: Thanks

PR #526, opened 2023-06-23
Implemented Delivery-drone task

OleksiiKiva:


app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot(Cargo):
+
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
```

app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
```
Bohdan-Salamakha, line 40: use typehint like: Optional[YOUR_VALUE_TYPE] = None

app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            coords: list[int] = None,
```
Bohdan-Salamakha, line 59: use typehint like: Optional[YOUR_VALUE_TYPE] = None

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #528, opened 2023-06-28
'Solution'

govorunenko:


app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
+
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None,
```
YukihiroSM, line 50: List of what? Add more descriptive definition, please. Also, if the element can be none - it should be marked as Optional - fix it everywhere.

YukihiroSM (changes requested):

app/main.py
govorunenko: I've fixed: Optional[list[int]].

YukihiroSM (approved):
Nicely done!

PR #529, opened 2023-06-28
Solution/yaroslav lysiuk

Quiet-Klirik:


app/main.py
```diff
@@ -1,5 +1,71 @@
...
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int, int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords else [0, 0]
```
Dimosphen1, line 18: Everything is okay with this line, just FYI you can use the `or` expression here and on line 43 as well to make a variable assignment. You can see a discussion of this approach in the [following thread](https://stackoverflow.com/questions/21566106/what-does-an-x-y-or-z-assignment-do-in-python).

Dimosphen1 (approved):
LGTM, nice work!

PR #537, opened 2023-07-03
Solution

3lobites:


app/main.py
```diff
@@ -2,4 +2,79 @@ class Cargo:
...
+        for key, values in kwargs.items():
+            if key == "name":
+                self.name = values
+            if key == "weight":
+                self.weight = values
+            if key == "coords":
+                if values is None:
+                    self.coords = [0, 0]
+                else:
+                    self.coords = values
```
olena-hul, line 30: This looks weird. Adjust your args to accept name, weight and coords explicitly

olena-hul: Fix in all classes

olena-hul (changes requested):

3lobites:
But I protest! I'm an artist, that's how I see it :)

app/main.py
```diff
@@ -2,4 +2,97 @@ class Cargo:
...
```
YukihiroSM, line 30: Why do you need to get the same variables from kwargs, if they are already processed previously? Also, kwargs will not contain `name`, `weight`, `coords`, because these variable names are already in use.

app/main.py
```diff
@@ -2,4 +2,97 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str = "",
+            weight: int = 0,
+            coords: list = None,
```
YukihiroSM, line 53: List of what? 
Add more detailed type hint like `list[SOME_DATA_TYPE]`

app/main.py
```diff
@@ -2,4 +2,97 @@ class Cargo:
...
+            self.coords = coords
+
+        if "max_load_weight" not in kwargs:
+            self.max_load_weight = 0
+        else:
+            self.max_load_weight = kwargs["max_load_weight"]
+        if "current_load" not in kwargs:
+            self.current_load = None
+        else:
+            self.current_load = kwargs["current_load"]
```
YukihiroSM, line 90: these variables should be delivered into __init__ in the same way as name, weight, coords.

YukihiroSM: Also, you can use ternary operator here.

YukihiroSM (changes requested):

3lobites:
Removed unnecessary kwargs in classes. The ternary operator is used - it's really half-hearted. 
Yesterday in the test "test_deliver_robot_hook_load_cargo_is_not_heavy" it seemed to me that "max_load_weight" and "current_load" are swapped :(

app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str = "",
+            weight: int = 0,
+            coords: list[int] = None
```

app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str = "",
+            weight: int = 0,
+            coords: list[int] = None
```

app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str = "",
+            weight: int = 0,
+            coords: list[int] = None,
+            max_load_weight: int = 0,
+            current_load: Cargo = None
```

app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = [0, 0, 0] if coords is None else coords
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None:
+            if cargo.weight <= self.max_load_weight:
```
Bohdan-Salamakha, line 66: unite these if conditions

Bohdan-Salamakha (changes requested):

3lobites:
Simplified the condition in function "hook_load". Used the type annotation "Optional"

Bohdan-Salamakha (approved):

PR #548, opened 2023-07-16
'Solution'

usKos:


app/main.py
```diff
@@ -1,5 +1,70 @@
...
 class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
```

app/main.py
```diff
@@ -1,5 +1,70 @@
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self, name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
```
Bohdan-Salamakha, line 40: use typehint like: Optional[YOUR_VALUE_TYPE] = None

app/main.py
```diff
@@ -1,5 +1,70 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None,
+                 max_load_weight: int = 0,
+                 current_load: Cargo = None) -> None:
```
Bohdan-Salamakha, line 58: use typehint like: Optional[YOUR_VALUE_TYPE] = None

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #552, opened 2023-07-19
Solution

Shadowtion:
Completed

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2,4 +2,66 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords else [0, 0]
+
+    def go_forward(self, step: int = 1) -> None:
+        self.coords[1] += step
+
+    def go_back(self, step: int = 1) -> None:
+        self.coords[1] -= step
+
+    def go_right(self, step: int = 1) -> None:
+        self.coords[0] += step
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
```

app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
```

app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: int,
+            coords: list[int] = None) -> None:
```

Bohdan-Salamakha (changes requested):

app/main.py
YukihiroSM: @Shadowtion Still not fixed. You should use Optional. Fix it, please.

app/main.py
YukihiroSM: @Shadowtion Still not fixed. You should use Optional. Fix it, please.

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[str] = None
```
YukihiroSM, line 11: coords are list of integers, not strings. Fix everywhere

YukihiroSM (changes requested):
Fix previous comment, please.

Shadowtion:
> Fix previous comment, please.

fixed

YukihiroSM (approved):
Well done! Good job!

PR #553, opened 2023-07-19
Delivery Drone

MishkaGAMII:


app/main.py
```diff
@@ -2,4 +2,78 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: tuple = None
```
```suggestion
                 coords: Optional[list[int]] = None)
```

vasylhnatiuk: change here and in other methods

vasylhnatiuk (approved):

PR #554, opened 2023-07-20
Solution2007

MichaelKyiv:


app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] = None) -> None:
```

vasylhnatiuk: add here and in all of code

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #555, opened 2023-07-20
solution

H1kki:


app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
Dimosphen1, line 10: Instead of using conditional variable assignment, you could consider using the `or` expression here. It would be more laconic and does not require using `None`. Read more about this [here](https://stackoverflow.com/a/21566107).

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Dimosphen1, line 29: You could expand type hinting here more, declaring that you expect to get the list of three integers for `coords`.

Dimosphen1 (approved):
LGTM, nice work! Several comments are left for future improvements.

PR #565, opened 2023-08-01
Solution

AlexPelevin99:


app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+class BaseRobot:
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: list[int] | None = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = [0, 0] if not coords else coords
```
Dimosphen1, line 15: Instead of conditional variable assignment, you could consider assigning attributes with the use of the `or` expression. You could read more about it [here](https://stackoverflow.com/questions/21566106/what-does-an-x-y-or-z-assignment-do-in-python) and use it in other cases if needed.

Dimosphen1 (approved):
LGTM, nice work! One comment is left for consideration.

PR #569, opened 2023-08-05
solution

EvgenMilokumov:


app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: float, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
```
vasylhnatiuk, line 10: will be better - self.coords = coords if coords else [0, 0]

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: float, coords: list = None) -> None:
```
```suggestion
    def __init__(self, name: str, weight: float, coords: Optional[list[int]] = None) -> None:
```

vasylhnatiuk: change everywhere

vasylhnatiuk (changes requested):

app/main.py
EvgenMilokumov: done, thanks

vasylhnatiuk (approved):

PR #571, opened 2023-08-09
Solution

eduardhabryd:


app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
```
Kroll410, line 7: [Indents ](https://stackoverflow.com/a/41591509):)

Kroll410 (changes requested):
All is good, but the indents

Kroll410 (approved):
Looks sharp now

PR #572, opened 2023-08-10
Solution

nyanplague:


app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+
+        if coords is None:
+            self.coords = [0, 0]
```
Kroll410, line 12: You can use ternary here instead

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+        max_load_weight: int,
+        current_load: Cargo,
+        coords: list = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
```
Kroll410, line 60: We are not assigning None to the current_load, so better to use `not`

Kroll410 (changes requested):

Kroll410:
Good job, but some changes need to be applied

Kroll410 (approved):

PR #573, opened 2023-08-10
'Solution'

lgSilay:


app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
```
Kroll410, line 7: Use [this ](https://stackoverflow.com/a/41591509)formatting (and in the future)

Kroll410 (changes requested):
Looks good, only formatting need to be applied

app/main.py
lgSilay: sorry for my formating. I'll try don't repeat this in a future

Kroll410 (approved):
GJ

PR #574, opened 2023-08-10
Implemented solution

RPodolets:


app/main.py
```diff
@@ -1,5 +1,77 @@
...
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        if coords:
+            self.coords = coords
```
Kroll410, line 19: You can do it with ternary operator

app/main.py
```diff
@@ -1,5 +1,77 @@
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        if not coords:
```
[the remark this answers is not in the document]
Kroll410: It's technically the same

app/main.py
```diff
@@ -1,5 +1,77 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Union[None, Cargo],
```
Kroll410, line 63: You can pass Optional typing here instead of Union

Kroll410 (changes requested):

PR #577, opened 2023-08-12
drones r ready

maxkatkalov:


app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: list[int] = None,
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = coords or [0, 0, 0]
```
Kroll410, line 41: It's a possible construction, but it adds implicitness to the code. Better use ternary

Kroll410 (changes requested):

Kroll410 (approved):

PR #588, opened 2023-08-14
Implement 3 classes with inheritance

o-davydova:


app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
Dimosphen1, line 16: You could consider using the `or` expression here and on line 45, you can read a discussion about it [here](https://stackoverflow.com/questions/21566106/what-does-an-x-y-or-z-assignment-do-in-python).

Dimosphen1 (approved):

app/main.py
o-davydova: Thanks for that :)

PR #612, opened 2023-08-22
solution py-delivery-drone

IvDyriv:


app/main.py
```diff
@@ -1,5 +1,73 @@
 class Cargo:
-    def __init__(self, weight: int) -> None:
+    def __init__(self,
+                 weight: int) -> None:
```
olena-hul, line 3: If arguments fit into 1 line, you can leave them in 1 line

app/main.py
```diff
@@ -1,5 +1,73 @@
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot(Cargo):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list | None = None) -> None:
```
olena-hul, line 11:
```suggestion
    def __init__(
        self,
        name: str,
        weight: int,
        coords: list | None = None
    ) -> None:
```

olena-hul: Fix everywhere

olena-hul (changes requested):

app/main.py
IvDyriv: Thank you. Okay

app/main.py
YukihiroSM: Not fixed.

YukihiroSM (changes requested):

app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot(Cargo):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list | None = None
```
Bohdan-Salamakha, line 11: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str]

app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list | None = None
```
Bohdan-Salamakha, line 38: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str]

app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: int,
+            coords: list = None
```
Bohdan-Salamakha, line 58: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str]
+
use typehint for None value like u used to did it before
coords: list | None = None

Bohdan-Salamakha (changes requested):

app/main.py
IvDyriv: Thanks for the tip. Done

app/main.py
IvDyriv: Thanks for the tip. Done

olena-hul (approved):

olena-hul:
Good job!

PR #616, opened 2023-08-29
Delivery drone solution

Filllick:


app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.coords = coords
+        if not self.coords:
```
vasylhnatiuk, line 33: same here

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+            self.coords = [0, 0, 0]
+        super().__init__(name, weight, self.coords)
```
vasylhnatiuk, line 35: better to use keywords arguments

vasylhnatiuk (changes requested):

Filllick:
Changed the coords assign and put keywords to super().__init__

app/main.py
```diff
@@ -2,4 +2,56 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] = None) -> None:
```

vasylhnatiuk: add everywhere

vasylhnatiuk (changes requested):

Filllick:
changed list to list[int] everywhere

app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None
+                 ) -> None:
+        self.coords = coords or [0, 0, 0]
+        super().__init__(name=name, weight=weight, coords=self.coords)
```
ArsenPidhoretskyi, line 39:
```suggestion
       coords = coords or [0, 0, 0]
        super().__init__(name=name, weight=weight, coords=coords)
```
Better to do like this

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):

Filllick:
in class FlyingRobot replaced self.coords with just coords

vasylhnatiuk (approved):

PR #622, opened 2023-09-02
solution

MykolaKrasulin:


app/main.py
```diff
@@ -2,4 +2,75 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot:
+
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
```

app/main.py
```diff
@@ -2,4 +2,75 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
```
Bohdan-Salamakha, line 43: use typehint like: Optional[YOUR_VALUE_TYPE] = None

app/main.py
```diff
@@ -2,4 +2,75 @@ class Cargo:
...
+        for _ in range(step):
+            self.coords[2] -= 1
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None,
```
Bohdan-Salamakha, line 62: use typehint like: Optional[YOUR_VALUE_TYPE] = None

Bohdan-Salamakha (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #624, opened 2023-09-02
Solution: Delivery Drone.

lobodInI:


app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
```

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
```

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None,
+            max_load_weight: int = 0,
+            current_load: Cargo = None
```

Bohdan-Salamakha (changes requested):

app/main.py
lobodInI: 👍

app/main.py
lobodInI: 👍

app/main.py
lobodInI: 👍

Bohdan-Salamakha (approved):

PR #626, opened 2023-09-03
Done

DimaPr7:


app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] = None) -> None:
```

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int,
+                 coords: list[int] = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords or [0, 0] or [0, 0, 0]
+
+    def go_forward(self, step: int = 1) -> list:
```
vasylhnatiuk, line 13:
```suggestion
    def go_forward(self, step: int = 1) -> list[int]:
```

vasylhnatiuk: add everywhere

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int,
+                 coords: list[int] = None) -> None:
+        self.coords = coords or [0, 0, 0]
+        super().__init__(name, weight, self.coords)
```
vasylhnatiuk, line 37:
```suggestion
        super().__init__(name=name, weight=weight, coords=self.coords)
```

vasylhnatiuk: better to use keywords arguments, change everywhere

vasylhnatiuk (changes requested):
last comment

vasylhnatiuk (approved):

PR #633, opened 2023-09-06
Solution

Lebwa1337:


app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            coords = [0, 0]
+        self.coords = coords
```
Dimosphen1, line 12: Use conditional assignment or [or expression](https://stackoverflow.com/questions/21566106/what-does-an-x-y-or-z-assignment-do-in-python) here to assign `self.coords`, avoid using multi-line code in simple operations that can be implemented in one line.

Dimosphen1 (changes requested):

Lebwa1337:
I made it in one line, but i don`t understand why i need to pass coords in __init__ here. Cuz i still need to redefine it due to adding a new Z axis

Dimosphen1 (approved):
LGTM, well done! Response comment is also left.

PR #635, opened 2023-09-09
solution

antonnech2309:


app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
pavlejviki, line 18:
```suggestion
            self.coords = coords or [0, 0]
```

pavlejviki, line 18: You can avoid using if/else

app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        super().__init__(name=name, weight=weight)
+        if coords is None:
+            self.coords = [0, 0, 0]
+        else:
+            self.coords = coords
```
pavlejviki, line 47: You can avoid using if/else

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #641, opened 2023-09-17
'Solution'

nsdNite:


app/main.py
```diff
@@ -2,4 +2,71 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
```
+
add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str]

app/main.py
```diff
@@ -2,4 +2,71 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
```
Bohdan-Salamakha, line 41: use typehint like: Optional[YOUR_VALUE_TYPE] = None
+
add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str]

app/main.py
```diff
@@ -2,4 +2,71 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo | None,
+            coords: list = None
```
Bohdan-Salamakha, line 60: use typehint like: Optional[YOUR_VALUE_TYPE] = None
+
add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str]

Bohdan-Salamakha (changes requested):

app/main.py
nsdNite: done, thanks for comments
used Optional[List[int]] = None for coords list where it is used, no other lists in code.

app/main.py
```diff
@@ -1,5 +1,75 @@
...
+            name: str,
+            weight: int,
+            coords: Optional[List[int]] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        if coords:
+            self.coords = coords
+        else:
+            self.coords = [0, 0]
```
Dimosphen1, line 21: You should prefer one style of the variable assignment in code, prefer picking up `or` as it is done on lines 65 and 46. Remove this `if ... else` condition and assign `self.coords` directly.

app/main.py
```diff
@@ -1,5 +1,75 @@
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: Optional[List[int]] = None
+    ) -> None:
+        coords = coords or [0, 0, 0]
```
Dimosphen1, line 46: You do not need to allocate `coords` variable here, pass this expression directly to the `super().__init__`

app/main.py
```diff
@@ -1,5 +1,75 @@
...
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo | None,
+            coords: Optional[List[int]] = None
+    ) -> None:
+        coords = coords or [0, 0, 0]
```
Dimosphen1, line 65: You do not need to allocate `coords` variable here, pass this expression directly to the `super().__init__`

Dimosphen1 (changes requested):

app/main.py
nsdNite: updated
thanks for comment

app/main.py
nsdNite: updated
thanks for comment

app/main.py
nsdNite: updated
thanks for comment

app/main.py
```diff
@@ -1,5 +1,70 @@
...
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo | None,
+            coords: Optional[List[int]] = None
+    ) -> None:
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+        super().__init__(name, weight, coords=coords or [0, 0, 0])
```
Dimosphen1, line 63: In case you pass the keyword arguments in the `__init__`, it would be better to pass other arguments with keys as well to be consistent and make sure that you do not produce ordering errors.

Dimosphen1 (approved):

app/main.py
nsdNite: Noted, thanks!

PR #642, opened 2023-09-18
implemented py-delivery-drone task

HlovaOleh:


app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
olena-hul, line 7: Give more detailed annotation for coords

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is not None and len(coords) == 2:
+            coords.append(0)
+        elif coords is None:
+            coords = [0, 0, 0]
```
olena-hul, line 33: Too complicated. It is enough to check for None only

app/main.py
olena-hul: Fix everywhere

olena-hul (changes requested):

PR #650, opened 2023-09-25
Implement 3 classes with inheritance

AndriiHomeniuk1:


app/main.py
```diff
@@ -1,5 +1,75 @@
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: Optional[List[int]] = None
+    ) -> None:
+        if coords is None:
```
vasylhnatiuk, line 43: write it in one line

app/main.py
```diff
@@ -1,5 +1,75 @@
...
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
```
vasylhnatiuk, line 45: use keyarguments

vasylhnatiuk (changes requested):

app/main.py
AndriiHomeniuk1: done)

vasylhnatiuk (approved):

PR #652, opened 2023-09-27
Solution

loplicat:


app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
+            current_load: Cargo | None,
+            coords: list = None,
+    ) -> None:
+        coords = coords if coords is not None else [0, 0, 0]
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if not self.current_load and \
```
Dimosphen1, line 65: Omit using the `\` operator, and write the expression inside `()` so there would be no need for it.

app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
Dimosphen1, line 15: You could omit using the `is not None` part, check for the boolean type of the `coords` would be enough. Besides that, you could use [or operator](https://stackoverflow.com/questions/21566106/what-does-an-x-y-or-z-assignment-do-in-python) to make variable assignment even more compact and expressive.

app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        coords = coords if coords is not None else [0, 0, 0]
```
Dimosphen1, line 40: Do not allocate a separate variable `coords`, perform this operation in a shortened form inside of the `super().__init__()` call.

app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo | None,
+            coords: list = None,
+    ) -> None:
+        coords = coords if coords is not None else [0, 0, 0]
```
Dimosphen1, line 59: Do not allocate a separate variable `coords`, perform this operation in a shortened form inside of the `super().__init__()` call.

app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
```
Dimosphen1, line 38: Indicate that `coords` is an optional argument here and on lines 11, 57 using type annotation.

Dimosphen1 (changes requested):

app/main.py
loplicat: Thanks for the tip, that's so convenient :)

Polyakiv-Andrey (approved):
Good job!

PR #658, opened 2023-10-04
Solution

kravetb:


app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = (0, 0)) -> None:
```
YukihiroSM, line 7: coords is list of what? Add more descriptive typehint like: List[str] if it's a list of strings. Also, the default value of coords is tuple. Are you sure that the typehint should be list?

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        max_load_weight: int,
+        name: str,
+        weight: int,
+        coords: list = (0, 0, 0),
+        current_load: None = None
```
YukihiroSM, line 51: current_load can be None, or int, right? Fix the typehint.

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1,5 +1,70 @@
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: tuple[int] = (0, 0)
```
EdAlekseiev, line 14: Why is `coords` a `tuple`? This must be a list

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,5 +1,78 @@
...
+        max_load_weight: int,
+        name: str,
+        weight: int,
+        coords: list[int] = None,
+        current_load: Union[None, "Cargo"] = None
+    ) -> None:
+        super().__init__(
+            name,
+            weight,
+            coords if coords is not None else [0, 0, 0]
```
ArsenPidhoretskyi, line 68: this will automatically do into super()

ArsenPidhoretskyi (changes requested):

app/main.py
kravetb: How I can do it?

app/main.py
```diff
@@ -1,5 +1,83 @@
...
+        coords: list[int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is not None:
+            self.coords = coords
+        elif isinstance(self, (FlyingRobot, DeliveryDrone)):
+            self.coords = [0, 0, 0]
+        else:
+            self.coords = [0, 0]
```
YukihiroSM, line 23: Base class shouldn't be dependent on it's child classes.

Better to do the following in init methods of child classes: 
 if coords is None:
            coords = [0, 0, 0]

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1,5 +1,85 @@
...
+        name: str,
+        weight: int,
+        coords: list[int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is not None:
+            self.coords = coords
+        else:
+            self.coords = [0, 0]
```
EdAlekseiev, line 21: You can rewrite this by using the ternary operator
```suggestion
        self.coords = coords if coords else [0, 0]
```

app/main.py
```diff
@@ -1,5 +1,85 @@
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: list[int] = None
```
EdAlekseiev, line 14: If a variable can be None you must show it a type hint
```suggestion
        coords: list[int] | None = None
```

app/main.py
```diff
@@ -1,5 +1,85 @@
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: list[int] = None
```
EdAlekseiev, line 44: The same comment as above

app/main.py
```diff
@@ -1,5 +1,85 @@
...
+        self.coords[2] -= point
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        max_load_weight: int,
+        name: str,
+        weight: int,
+        coords: list[int] = None,
```
EdAlekseiev, line 67: The same comment as above

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #660, opened 2023-10-05
Solution

Naz-iv:


Daniil-Pankieiev:
Looks okay, but im not sure about this: 
super().__init__(name,
                         weight,
                         coords if coords is not None else [0, 0, 0])

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+
+class BaseRobot:
+
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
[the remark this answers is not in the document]
Naz-iv: didn't think about it, thanks for advice

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+                 current_load: int,
+                 coords: list = None) -> None:
+        super().__init__(name,
+                         weight,
+                         coords if coords is not None else [0, 0, 0])
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if (self.current_load is None
```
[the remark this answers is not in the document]
Naz-iv: corrected, thanks

Wonsky1 (approved):

app/main.py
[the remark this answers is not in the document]
fsocie7y: u forgot fix that))

fsocie7y (changes requested):

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int,
+                 coords: list = None) -> None:
+        super().__init__(name,
+                         weight,
+                         coords if coords else [0, 0, 0])
```
fsocie7y, line 59: no need to make this check here, it is already checked in parent class

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int,
```
fsocie7y, line 55: and fix type annotation for current_load
it takes Cargo object or None
but not int))

fsocie7y (changes requested):

app/main.py
Naz-iv: but parent class has two axis [0, 0], but drone needs to have three - [0, 0, 0]

app/main.py
Naz-iv: updated

fsocie7y (approved):
well done ^_^

PR #663, opened 2023-10-07
Solution for py_delivery_drone

wataschi:


app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot(Cargo):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(weight)
+        self.name = name
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
viktoria-rybenchuk, line 13: use ternary operator here. Fix in all places

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+        super().__init__(name, weight)
+        self.max_load_weight = max_load_weight
+        self. current_load = current_load
+        if coords is None:
+            self.coords = [0, 0, 0]
+        else:
+            self.coords = coords
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
```
viktoria-rybenchuk, line 64: not need check explicitly for None. You can check `if not self.current_load`

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot(Cargo):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(weight)
+        self.name = name
+        self.coords = [0, 0] if coords is None else coords
```
viktoria-rybenchuk, line 10:
```suggestion
        self.coords = [0, 0] if coords else coords
```

viktoria-rybenchuk: fix in all places

viktoria-rybenchuk (changes requested):

app/main.py
wataschi: A little differently:
self.coords = [0, 0] if coords else coords

wataschi: a little differently:

pavlejviki (approved):
Well done!

PR #670, opened 2023-10-08
Solution

kostomeister:


D1st3f (approved):
GJ!

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None,
+            max_load_weight: int = 0,
+            current_load: int = None
+    ) -> None:
+        if coords is None:
```
fsocie7y, line 52: no need to make this check here, it is already checked in parent class

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
```
fsocie7y, line 8: Use a single composite `if` statement
in similar case too

fsocie7y (changes requested):

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if not coords:
+            coords = [0, 0]
```
fsocie7y, line 9: u can do it in one line))

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if not coords:
```
fsocie7y, line 32: here too in one line
and after this change u can move it in super().__init__()

fsocie7y (changes requested):

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
...
+        super().__init__(name, weight, coords)
+        self.current_load = current_load
+        self.max_load_weight = max_load_weight
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load:
+            return
+
+        if cargo.weight <= self.max_load_weight:
+            self.current_load = cargo
```
fsocie7y, line 57: no need return here
u can use just one condition)

fsocie7y (changes requested):

app/main.py
```diff
@@ -2,4 +2,56 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None,
+            max_load_weight: int = 0,
+            current_load: int = None
```
fsocie7y, line 46: fix type annotation for current_load
it should take Cargo object or None
but not int)))

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #682, opened 2023-10-10
Solution

OlehShumov:


app/main.py
```diff
@@ -1,5 +1,69 @@
+from typing import List, Optional
+
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: Optional[List[int]] = None
```
LiudmylaKulinchenko, line 8: Use list[int] | None instead,  fix everywhere

app/main.py
```diff
@@ -1,5 +1,69 @@
...
+                 ) -> None:
```
LiudmylaKulinchenko, line 9: Fix formatting, fix everywhere

LiudmylaKulinchenko (changes requested):

fsocie7y (approved):
well done ^_^

PR #684, opened 2023-10-10
Solution

Shatrovskyi:


MiskoRuslan (approved):
interesting solution with coords

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: None,
```
fsocie7y, line 48: You use two different data types for these arguments, so use the correct type annotation current_load: Cargo | None = None. Fix in all other similar cases

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = [0, 0, 0] if coords is None else coords
```
fsocie7y, line 33: you can move it to the super

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #686, opened 2023-10-10
Solution

MiskoRuslan:


RabenkoYevhenii (approved):

app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int = 0,
+                 current_load: int = None,
+                 coords: list[int] = None) -> None:
+
+        super().__init__(name,
+                         weight,
+                         coords=coords if coords is not None else [0, 0, 0])
```
fsocie7y, line 59: no need to make this check here, it is already checked in parent class

app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
fsocie7y, line 14: better use if coords else [0, 0]) its prevents empty lists
in similar case too)

fsocie7y (changes requested):

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+
+        super().__init__(name,
+                         weight,
+                         coords=coords if coords is not None else [0, 0, 0])
```
fsocie7y, line 40: same as i wrote earlier about if coords))

fsocie7y (changes requested):

app/main.py
MiskoRuslan: done c :

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+                 coords: list[int] = None) -> None:
+
+        super().__init__(name, weight, coords)
+
+        self.max_load_weight = max_load_weight
+        self. current_load = current_load
+
+    def hook_load(self, other: Cargo) -> None:
+        if self.current_load is not None:
+            return None
```
fsocie7y, line 64: no need return here

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+        if self.max_load_weight >= other.weight:
```
fsocie7y, line 65: and u can do it in one condition)

fsocie7y (changes requested):

app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int = 0,
+                 current_load: int = None,
```
fsocie7y, line 54: fix type annotation for current_load
it should take Cargo object or None
but not int)))

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #691, opened 2023-10-10
Solution Delivery drone

Oxbay:


fedorov-s-od (approved):
In my humble opinion it looks good )

fsocie7y (approved):
good, but in the future try to do type annotations without quotes)

PR #703, opened 2023-10-15
'Solution

oleksandr-pupchenko:


app/main.py
```diff
@@ -1,5 +1,58 @@
...
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(self, name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: Cargo,
```
LiudmylaKulinchenko, line 47: it also could be None

app/main.py
```diff
@@ -1,5 +1,58 @@
...
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
LiudmylaKulinchenko, line 32: coords also could be None. Because if you get coords = None in DeliveryDrone, when u use `super`, you pass None value for coords here. So fix annotatons everywhere

LiudmylaKulinchenko (changes requested):

fsocie7y (approved):
well done ^_^

PR #705, opened 2023-10-16
Solution

DanSheremeta:
Solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2,4 +2,70 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+        if self.coords is None:
+            self.coords = [0, 0]
+
+    def go_forward(self, step: int = 1) -> None:
+        self.coords[1] += step
+
+    def go_back(self, step: int = 1) -> None:
+        self.coords[1] -= step
+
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+        if self.coords is None:
+            self.coords = [0, 0]
```
EdAlekseiev, line 17: Use the ternary operator here

app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        if self.coords == [0, 0]:
+            self.coords = [0, 0, 0]
```
EdAlekseiev, line 44: I think, that you need to check coords are None before super

app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
```
EdAlekseiev, line 11: If argument can be None you have to show in type hint. Fix here and in other functions

EdAlekseiev, line 11:
```suggestion
            coords: list[int] | None = None
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #707, opened 2023-10-17
Solution amazon

hbilous:


app/main.py
```diff
@@ -2,4 +2,54 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
YukihiroSM, line 7: If the default value of some variable is None - better to use `Optional[list]` or `list | None` syntax. 
Also add more descriptive typehint for list. Example: is it's a list of strings: `list[str]`

YukihiroSM: Fix it in other functions

YukihiroSM (changes requested):

YukihiroSM (changes requested):
If the default value of some variable is None - better to use Optional[list] or list | None syntax.

hbilous:
> If the default value of some variable is None - better to use Optional[list] or list | None syntax.

image
Sorry, I don't understand what do you want from me

EdAlekseiev (approved):
Good job

PR #714, opened 2023-10-22
Solution

CrazyHilly:


app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
+            name: str,
+            weight: int,
+            coords: list[int] | None = None
+    ) -> None:
+
+        self.name = name
+        self.weight = weight
+        if not coords:
+            coords = [0, 0]
+        self.coords = coords[:2]
```
ArsenPidhoretskyi, line 18: Write this inline

app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] | None = None
+    ) -> None:
+
+        super().__init__(name, weight, coords)
+        z_coord = 0 if not coords else coords[2]
```
ArsenPidhoretskyi, line 45:
```suggestion
        super().__init__(name, weight, coords or [0, 0, 0])
```
z_coord is redundant

ArsenPidhoretskyi (changes requested):

app/main.py
CrazyHilly: [0, 0, 0] will be redundant here since the coords are passed to BaseRobot init method where they are truncated anyway. 
But I did get rid of z_coord by condensing the next two lines into one.

EdAlekseiev (approved):
Good job

PR #717, opened 2023-10-25
Solution

Artem-user-19:


app/main.py
```diff
@@ -1,5 +1,72 @@
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: Optional[List[int]] = None
+    ) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
```
viktoria-rybenchuk, line 44: Use the ternary operator here

app/main.py
```diff
@@ -1,5 +1,72 @@
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: Optional[List[int]] = None
```
viktoria-rybenchuk, line 14:
```suggestion
            coords: Optional[List[int]]   | None = None
```

viktoria-rybenchuk: Fix in all places

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #723, opened 2023-10-29
Solution

OLENA-KALITSINSKA:


app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: int,
```
viktoria-rybenchuk, line 56: write correct type annotation for `current_load`

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int | float] = None
```
Dimosphen1, line 38: You could indicate that arguments are optional in type annotations in this way:
```suggestion
            coords: list[int | float] | None = None
```
Apply this change to type annotation in places where you have optional arguments.

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+    ) -> None:
+        super().__init__(name, weight)
+        self.coords = coords or [0, 0, 0]
```
Dimosphen1, line 41: Do not redefine `self.coords` after the `__init__`, pass the correct value inside of the `super().__init__` directly, e.g.:
```suggestion
        super().__init__(name, weight, coords or [0, 0, 0])
```

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #736, opened 2023-11-07
Solution

lyaskov:


app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int,
+                 coords: list = None) -> None:
+
+        if coords is None:
+            coords = [0, 0, 0]
```
ArsenPidhoretskyi, line 52: This lines redundant it already be done into super class

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
EdAlekseiev, line 7: If argument can be `None` you have to show this fact in type hint. Also, you must specify inner type of list
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] | None = None) -> None:
```

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
EdAlekseiev, line 31: The same comment as above

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int,
+                 coords: list = None) -> None:
```
EdAlekseiev, line 49: The same comment as above

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #738, opened 2023-11-09
Solution

Liliyakhu:


app/main.py
```diff
@@ -1,5 +1,81 @@
...
+            name: str,
+            weight: int,
+            coords: Optional[List[int]] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
Dimosphen1, line 21: Assign a value to `self.coords` in one line, you could do it using the `if ... else` construction or with the use of [or assignment](https://stackoverflow.com/questions/21566106/what-does-an-x-y-or-z-assignment-do-in-python).

app/main.py
```diff
@@ -1,5 +1,81 @@
...
+            self,
+            name: str,
+            weight: int,
+            coords: Optional[List[int]] = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
+        else:
+            self.coords = coords
```
Dimosphen1, line 50: There is no need to redefine the variable after its initialization, rewrite the `__init__` to avoid that: 
```suggestion
        super().__init__(name, weight, coords or [0, 0, 0])
```

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,5 +1,74 @@
...
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo | None,
+            coords: Optional[List[int]] = None
+    ) -> None:
+        super().__init__(
+            name=name,
+            weight=weight,
+            coords=coords)
```
Polyakiv-Andrey, line 64: The closing bracket is also written on a new line

Polyakiv-Andrey (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #740, opened 2023-11-11
Solution

SashaTarnavsky:


app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> None:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
NikOneZ1, line 29: If argument can be None you have to show this fact in type hint. Also, you must specify inner type of list
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] | None = None) -> None:
```

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
NikOneZ1, line 7: If argument can be None you have to show this fact in type hint. Also, you must specify inner type of list

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None,
+            max_load_weight: int = 0,
+            current_load: bool = None
```
NikOneZ1, line 48: If argument can be None you have to show this fact in type hint. Also, you must specify inner type of list

NikOneZ1 (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #744, opened 2023-11-12
Solution

Vityk-4W:


ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
```
EdAlekseiev, line 11: Ifan argument can be `None`, you must show it in a type hint
```suggestion
            coords: list[int] | None = None
```

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = ([0, 0] if coords is None else coords)
```
EdAlekseiev, line 15:
```suggestion
        self.coords = [0, 0] if coords is None else coords
```

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
```
EdAlekseiev, line 38: The same comment as above

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = ([0, 0, 0] if coords is None else coords)
```
EdAlekseiev, line 41:
```suggestion
        self.coords = [0, 0, 0] if coords is None else coords
```

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo,
+            coords: list[int] = None
```
EdAlekseiev, line 57: The same comment as above

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #750, opened 2023-11-13
'Solution'

alex329657:


app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
+        if step >= 0:
+            self.coords[1] += step
+        else:
+            self.coords[1] -= step
+
+    def go_back(self, step: int = 1) -> None:
+        if step >= 0:
+            self.coords[1] -= step
+        else:
+            self.coords[1] += step
```
ArsenPidhoretskyi, line 24: Why? It is not correct go back must be do -= always, if step will be -3 why we should also go_back for me it is go forward

ArsenPidhoretskyi: Fix in all places

app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
+            self.coords[2] -= step
+        else:
+            self.coords[2] += step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self, name: str, weight: int, coords: list = None,
+                 max_load_weight: int = 0, current_load: Cargo = None) -> None:
+        if not coords:
+            coords = [0, 0, 0]
```
ArsenPidhoretskyi, line 65: This is redundant, this will done into super class

ArsenPidhoretskyi (changes requested):

app/main.py
alex329657: > Why? It is not correct go back must be do -= always, if step will be -3 why we should also go_back for me it is go forward

The task description does not indicate that the values ​​are always equal to or greater than zero. And the drone control can be reversed) Okay, I’ll change it.

app/main.py
```diff
@@ -2,4 +2,54 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if not coords:
+            coords = [0, 0]
+        self.coords = coords
```
vasylhnatiuk, line 12:
```suggestion
        self.coords = coords or [0, 0]
```

app/main.py
```diff
@@ -2,4 +2,54 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if not coords:
```
vasylhnatiuk, line 10: delete this, you can simplify it

app/main.py
```diff
@@ -2,4 +2,54 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] = None) -> None:
```

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -2,4 +2,54 @@ class Cargo:
...
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if not coords:
```
vasylhnatiuk, line 32: the same here

app/main.py
```diff
@@ -2,4 +2,54 @@ class Cargo:
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
vasylhnatiuk, line 31: same here

vasylhnatiuk (changes requested):

app/main.py
vasylhnatiuk: do it

app/main.py
```diff
@@ -2,4 +2,54 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int,
+                 coords: list[int] = None) -> None:
+        if not coords:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
```
vasylhnatiuk, line 34:
```suggestion
        super().__init__(name, weight, coords or [0, 0, 0])
```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #753, opened 2023-11-14
Implemented drone classes

Dikey945:


app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
...
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        if coords is None:
+            coords = [0, 0]
```
LiudmylaKulinchenko, line 14: rewrite in one line using ternary operator. Fix everywhere
It will be more readable in this case!

app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
...
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
```
LiudmylaKulinchenko, line 44: then here you could place it directly in super

app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
...
+            coords: list = None,
+    ) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
```
LiudmylaKulinchenko, line 69: you can also use `if not self.current_load`

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = [0, 0] if not coords else coords
```
LiudmylaKulinchenko, line 15: probably will be more clear `coords if coords else [0, 0]`
But it's also fine, as for me 😋

LiudmylaKulinchenko (approved):
Well done 🌺

PR #758, opened 2023-11-21
Solution

KwintDimon:


app/main.py
```diff
@@ -1,5 +1,72 @@
...
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: float,
+            coords: list[float] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
viktoria-rybenchuk, line 13: you don't need explicitly check `if coords is not None`

app/main.py
```diff
@@ -1,5 +1,72 @@
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: float,
+            coords: list[float] = None
+    ) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
```
viktoria-rybenchuk, line 39: use ternary operator here as well

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: float,
+            coords: list[float] = None
+    ) -> None:
+        coords = coords if coords else [0, 0, 0]
+        super().__init__(name, weight, coords)
```
Dimosphen1, line 39: As an option, you could consider using the following syntax:
```suggestion
        super().__init__(name, weight, coords or [0, 0, 0])
```

Dimosphen1 (approved):

PR #766, opened 2023-11-27
Solution

rinotokun:


app/main.py
```diff
@@ -1,5 +1,66 @@
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords:
+            self.coords = coords
+        else:
+            self.coords = [0, 0]
```
Dimosphen1, line 16: Do not write self.coords using multiline expression, you could implement the same logic using if else
```suggestion
        self.coords = coords if coords else [0, 0]
```
or using `or` 
```suggestion
        self.coords = coords or [0, 0]
```

app/main.py
```diff
@@ -1,5 +1,66 @@
...
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords:
+            super().__init__(name, weight, coords)
+        else:
+            super().__init__(name, weight, coords=[0, 0, 0])
```
Dimosphen1, line 39: There is no need to implement two alternative `__init__` statements, reduce their number to one:
```suggestion
        super().__init__(name, weight, coords or [0, 0, 0])
```

Dimosphen1 (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #777, opened 2023-12-05
Solution

OlgaRozd:


OlgaRozd:
corrected

app/main.py
```diff
@@ -1,5 +1,82 @@
...
+
+
 class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Dimosphen1, line 10: Place all arguments inside of the `__init__` on a new line to improve the code readability. Make this change to the `__init__` for all classes.

app/main.py
```diff
@@ -1,5 +1,82 @@
...
+        return self.coords
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 coords: list = None,
+                 current_load: int = None) -> None:
```
Dimosphen1, line 70: Place each argument on a new line, open and closing parenthesis should not have arguments on the same line with them. It is a conventional approach to form `__init__` with type annotations.

app/main.py
```diff
@@ -1,5 +1,82 @@
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0]
+        self.name = name
+        self.weight = weight
+        self.coords = coords
```
Dimosphen1, line 15: Do not use multi-line if condition to define `coords` make it in one line as this
```suggestion
        self.name = name
        self.weight = weight
        self.coords = coords or [0, 0]
```
or as this 
```suggestion
        self.name = name
        self.weight = weight
        self.coords = coords if coords else [0, 0]
```
Make this change to all places where applicable.

app/main.py
```diff
@@ -1,5 +1,82 @@
...
+
+    def go_forward(self, step: int = 1) -> list:
+        x, y, *z = self.coords
+        y += step
+        self.coords = [x, y, *z]
+        return self.coords
```
Dimosphen1, line 21: You could simplify the logic by making the in-place list modification, e.g.
```suggestion
        self.coords[1] += step
```
Apply this change to all methods that modify coordinates for all classes.

app/main.py
```diff
@@ -1,5 +1,82 @@
...
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 coords: list = None,
+                 current_load: int = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
```
Dimosphen1, line 72: You could pass `coords` in the form you obtained them, the necessary work would be done in the parent class.

app/main.py
```diff
@@ -1,5 +1,82 @@
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
```
Dimosphen1, line 49: You could use `or` to define the argument that you pass to the `super().__init__`
```suggestion
        super().__init__(name, weight, coords or [0, 0, 0])
```

Dimosphen1 (changes requested):

pavlejviki (changes requested):
Please fix all the previous comments

PR #781, opened 2023-12-09
Solution

halyna-baklanova:


app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = [0, 0, 0] if coords is None else self.coords
```
Dimosphen1, line 42: You should define the `coords` before passing them to the `super().__init__()`, otherwise you risk populating `[0, 0]` from the parent class here, because `[0, 0]` is truthy.
```suggestion
        super().__init__(name, weight, coords or [0, 0, 0])
```

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+        self.coords = [0, 0] if coords is None else self.coords
```
Dimosphen1, line 16: There is no need to have to line of `self.coords` definition, you could reduce the code: 
```suggestion
        self.coords = coords if coords else [0, 0]
```

Dimosphen1 (changes requested):

pavlejviki (approved):
GJ!

PR #784, opened 2023-12-12
Solution

DanuloM:


app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0]
```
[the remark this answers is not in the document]
DanuloM: image
i dont really understand how it supposed to work, could you help me?

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            max_load_weight: int,
+            current_load: int,
+            name: str,
+            weight: int,
+            coords: list[3] = None
```
Dimosphen1, line 50: You expect a `list` of `int`, change your type annotation in places where needed.

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
+    ) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        super().__init__(name=name, weight=weight, coords=coords)
```
Dimosphen1, line 54: There is no need to redefine `coords` in this place, all the related work should be done in the parent class. You could pass `coords` to the `super().__init__()` as it is.

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Dimosphen1, line 7: Place all arguments in the `__init__` on a new line to improve the readability.

app/main.py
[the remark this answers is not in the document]
Dimosphen1: This is an application of the `or` expression that is used on assignment. It allows to reduce the complexity and make code more concise and explicit. You can read more about it in [this thread](https://stackoverflow.com/questions/21566106/what-does-an-x-y-or-z-assignment-do-in-python). Apply it to all places where relevant. You could use `or` to dynamically define passed values in the `super().__init__()` as well.

Dimosphen1 (changes requested):

vasylhnatiuk (approved):

PR #786, opened 2023-12-12
Solution

Artkom98:


app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
 
+
 # write your code here
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            coords = [0, 0]
+        self.coords = coords
```
```suggestion
        self.coords = coords if coords is not None else [0, 0]
```

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
+
 # write your code here
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
EdAlekseiev, line 8: If argument can be None you must show it type hint. Also, you must specify inner type of list. Fix everywhere.
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] | None = None) -> None:
```

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 current_load: None,
+                 max_load_weight: int,
+                 coords: list[int] | None = None) -> None:
+        super().__init__(name, weight, coords)
+
+        self.current_load = current_load
```
YukihiroSM, line 58: According to your typehint current load can be only None. Then why do you need this parameter in init? 
Should current_load be only None, or only it's default value should be None?

YukihiroSM (changes requested):

EdAlekseiev (approved):
Good job

PR #789, opened 2023-12-18
'Solution'

yablonskyiv:


EdAlekseiev (changes requested):
Fix tests

EdAlekseiev (changes requested):
Fix tests

app/main.py
```diff
@@ -2,4 +2,52 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
EdAlekseiev, line 7: If an argument can be None, you must show it in type hint. Also, specify inner type of list
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] | None = None) -> None:
```

app/main.py
```diff
@@ -2,4 +2,52 @@ class Cargo:
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
EdAlekseiev, line 29: The same comment as above

app/main.py
```diff
@@ -2,4 +2,52 @@ class Cargo:
...
+    def go_up(self, step: int = 1) -> None:
+        self.coords[2] += step
+
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self, name: str, weight: int, coords: list = None,
+                 max_load_weight: int = 0, current_load: None = None) -> None:
```
EdAlekseiev, line 43: the same comment as above

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
...
+
+    def go_up(self, step: int = 1) -> None:
+        self.coords[2] += step
+
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self, name: str, weight: int, coords: list = None,
```
YukihiroSM, line 44:
```suggestion
    def __init__(self, name: str, weight: int, coords: list | None = None,
```

YukihiroSM (approved):
That's the only issue. Else looks good!

PR #796, opened 2023-12-22
Solution

Zonda001:


app/main.py
```diff
@@ -2,4 +2,52 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: float, coords: int = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
vsmutok, line 10:
```suggestion
        self.coords = coords or [0, 0]
```
better to use `or`

vsmutok (changes requested):

app/main.py
Zonda001: Ok, thanks

vsmutok (approved):
Well Done!

PR #799, opened 2023-12-25
Solution

xrkxdd:


app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int]= None) -> None:
```

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+        self.name = name
+        self.weight = weight
+        if coords is None:
```
vasylhnatiuk, line 10: try to write in oneline

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self, name: str,
+            weight: int,
+            coords: list = None
```
```suggestion
            coords: list[int] = None
```

vasylhnatiuk: do it everywhere

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #821, opened 2024-01-10
Solution

DenZatolokin:


app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
EdAlekseiev, line 7: If argument can be `None` you should show in type hint. Fix everywhere
```suggestion
    def __init__(self, name: str, weight: int, coords: list | None = None) -> None:
```

EdAlekseiev (changes requested):

NikOneZ1 (approved):
Well done!

PR #829, opened 2024-01-15
Solution

barabarinov:


app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: list[int, int] | list[int, int, int] | None = None
+    ) -> None:
```
msymonovych, line 12: Fix indentation here

app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        max_load_weight: int,
+        coords: list[int, int, int] = None,
+        current_load: int = None,
+    ) -> None:
```
msymonovych, line 61: You should fix indentation here as well

msymonovych (changes requested):

app/main.py
barabarinov: > Fix indentation here

I did it knowingly. That is black style which i like!
https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html

PR #830, opened 2024-01-15
Solution

Yevhen23tk:


app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
```
EdAlekseiev, line 10: If an argument can be `None` you must show it in type hint. Fix everywhere, please 
```suggestion
                 coords: list[int] | None = None) -> None:
```

EdAlekseiev (changes requested):

app/main.py
Yevhen23tk: ok, accepted

EdAlekseiev (approved):
Good job

PR #833, opened 2024-01-16
Solution

oksankakot:


app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
+
+    def __init__(self,
+                 name: str,
+                 weight: float,
+                 coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
+        else:
+            self.coords = coords
```
mdubyna, line 45: we need to change the sequence, if we look at the task we will see that the __init__ of the parent method should contain the default value [0, 0, 0]
image

mdubyna (changes requested):

app/main.py
oksankakot: Thanks, good remark

app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
+    def __init__(self,
+                 name: str,
+                 weight: float,
+                 coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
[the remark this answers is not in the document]
oksankakot: Done

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+
+    def __init__(self,
+                 name: str,
+                 weight: float,
+                 coords: list = None) -> None:
```
NEkropo1, line 11: are you sure init params could be only type of list?)

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            coords = [0, 0]
+        self.coords = coords
```
NEkropo1, line 16: make this in 1 lane

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(self,
+                 name: str,
+                 weight: float,
+                 max_load_weight: int,
+                 current_load: int,
```
NEkropo1, line 57: current load is int?)

NEkropo1 (changes requested):

app/main.py
```diff
@@ -1,5 +1,69 @@
...
+
+class BaseRobot:
+
+    def __init__(self,
+                 name: str,
+                 weight: float,
+                 coords: list[int] = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = [0, 0] if coords is None else coords
```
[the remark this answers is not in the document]
oksankakot: I got it, thanks

app/main.py
```diff
@@ -1,5 +1,69 @@
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self,
+                 name: str,
+                 weight: float,
+                 coords: list[int] = None) -> None:
```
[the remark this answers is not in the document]
oksankakot: Thanks.

PR #845, opened 2024-01-17
Solution

Serejka21:


Dobryk12 (approved):

app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list | None = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = [0, 0] if coords else coords
```
NEkropo1:
```suggestion
coords or [0, 0]
```

app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
...
+                 max_load_weight: int = 0,
+                 current_load: Cargo = None) -> None:
+
+        coords = coords if coords is not None else [0, 0, 0]
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
```
NEkropo1, line 61: you don't need to check if it's None in python, simply check if it is or if no it is. like:
```suggestion
        if not self.current_load and cargo.weight <= self.max_load_weight:
```

NEkropo1 (changes requested):

app/main.py
Serejka21: > 

I think you are checking the wrong commit, the implementation looks different in "fixed annotation 2"

Serejka21: Ok i will use it next time but now i would left it

PR #847, opened 2024-01-17
"Create 3 classes with inheritance'

achernyshovv:
Added BaseRobot class with methods for movement and information retrieval.
Implemented FlyingRobot as a subclass of BaseRobot, introducing vertical movement methods.
Created DeliveryDrone as a subclass of FlyingRobot, including methods to hook and unhook cargo.
Defined the Cargo class to represent cargo weight.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,5 +1,76 @@
+from __future__ import annotations
+from typing import Optional, List
+
+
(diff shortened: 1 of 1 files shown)


MaxymChyncha (approved):
Good job!

PR #851, opened 2024-01-17
Solution

kstorozhenko:


app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot(Cargo):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None) -> None:
```
VladKyianytsia, line 10: use more descriptive annotations: coords: list[int] | None

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+        super().__init__(weight)
```
VladKyianytsia, line 11: Cargo is not parent class for BaseRobot, Cargo is what these robots will deliver). So you are not supposed to use super() here

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+        self.name = name
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
VladKyianytsia, line 16: it's a good case for ternary

VladKyianytsia (changes requested):

app/main.py
NEkropo1: yup, use it

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = [0, 0] if coords is None else coords
```
NEkropo1, line 13:
```suggestion
        self.coords = coords or [0, 0]
```

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str,
+                 weight: int,
+                 coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
```
NEkropo1, line 36: same here

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: Cargo = None,
+                 coords: list[int] | None = None) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if (self.current_load is None
```
NEkropo1, line 58: if not load. It
```suggestion
        if not self.current_load
```

NEkropo1 (changes requested):

PR #853, opened 2024-01-18
Solution

mykytaso:


app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: [list] = None) -> None:
```
```suggestion
    def __init__(self, name: str, weight: int, coords: [int] = None) -> None:
```

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+        if coords is None:
```
vasylhnatiuk, line 8: try to do it in one line

vasylhnatiuk (changes requested):

app/main.py
mykyta-so: Unfortunately, I don't know how :(

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: int,
+            coords: [list] = None
```
vasylhnatiuk, line 48: fix here also

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #854, opened 2024-01-18
solution

eugenevsn:


app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: float, coords: list = None) -> None:
```
EdAlekseiev, line 7: If argument can be `None`, you must show it in type hint. Fix everywhere
```suggestion
    def __init__(self, name: str, weight: float, coords: list | None = None) -> None:
```

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,5 +1,61 @@
...
+        self.coords[2] += step
+
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self, name: str, weight: float, max_load_weight: float,
+            current_load: "Cargo" = None, coords: list | None = None) -> None:
```
EdAlekseiev, line 49:
```suggestion
            current_load: Cargo | None = None, coords: list | None = None) -> None:
```

EdAlekseiev (changes requested):

NikOneZ1 (approved):
Nicely done!

PR #857, opened 2024-01-25
Solution

AlexGrytsai:


app/main.py
```diff
@@ -1,5 +1,75 @@
...
+
+
 class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
EdAlekseiev, line 10: If argument can be None, you must show it in type hint. Also provide inner type of list. Fix everywhere
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] | None = None) -> None:
```

app/main.py
```diff
@@ -1,5 +1,75 @@
...
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
EdAlekseiev, line 16: Use ternary operator here

app/main.py
```diff
@@ -1,5 +1,75 @@
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
+        else:
+            self.coords = coords
```
EdAlekseiev, line 44: The same comment as above

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #858, opened 2024-01-26
Solution

ziriael-absten:


app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
EdAlekseiev, line 7: If an argument can be `None` you must show it in a type hint. Fix everywhere.
Also provide inner type of list
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] | None = None) -> None:
```

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+        self.name = name
+        self.weight = weight
+        if coords:
+            self.coords = coords
+        else:
+            self.coords = [0, 0]
```

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords:
+            coords = coords
+        else:
+            coords = [0, 0, 0]
```
EdAlekseiev, line 36: The same comment as above

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #859, opened 2024-01-26
Solution

serji13:


app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
YukihiroSM, line 7:
```suggestion
    def __init__(self, name: str, weight: int, coords: Optional[list] = None) -> None:
```

YukihiroSM: If the default value is None - parameter should be marked as Optional
Fix it everywhere

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+            current_load: int | None = None
+
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, other: Cargo) -> None:
+        if not self.current_load:
+            if other.weight <= self.max_load_weight:
```
YukihiroSM, line 62:
```suggestion
        if not self.current_load and other.weight <= self.max_load_weight:
```

YukihiroSM (changes requested):

YukihiroSM (approved):
Good job!

PR #860, opened 2024-01-27
Solution

EugeneProtsenko:


app/main.py
```diff
@@ -1,5 +1,62 @@
...
+
+
 class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
YukihiroSM, line 10:
```suggestion
    def __init__(self, name: str, weight: int, coords: Optional[list[int]] = None) -> None:
```

If the default value is None - parameter should be marked as Optional 
Also, it's a good thing to add more detailed typehint like list[int] to show, that you want to receive only list of integers. 
Fix it everywhere

YukihiroSM (changes requested):

app/main.py
EugeneProtsenko: Ok, got it

YukihiroSM (approved):
Nicely done!

PR #868, opened 2024-01-29
solution

Eugen-Veklich:


app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str,
+                 weight: int, coords: list[int] = None) -> None:
```
EdAlekseiev, line 8: If argument can be `None` you must show it in type hint. Fix everywhere
```suggestion
                 weight: int, coords: list[int] | None = None) -> None:
```

EdAlekseiev (changes requested):

NikOneZ1 (approved):

PR #869, opened 2024-01-30
solution

prev1ew:


app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int | float,
+            coords: list[int] | None = None,
+            max_load_weight: int = 0,
+            current_load: Cargo = None
+    ) -> None:
+        super().__init__(name, weight, coords or [0, 0, 0])
```
Dimosphen1, line 58: There is no need to perform or expression assignment here because the same check is performed in the parent class. Remove it to avoid the code duplication.
```suggestion
        super().__init__(name, weight, coords)
```

Dimosphen1 (changes requested):
Several minor changes were requested.

Dimosphen1 (approved):
LGTM, great job!

PR #874, opened 2024-02-03
Solution

MilArtem78:


app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int, int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
pavlejviki, line 15:
```suggestion
        self.coords = coords if coords  else [0, 0]
```

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int, int, int] = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = coords if coords is not None else [0, 0, 0]
```
pavlejviki, line 41:
```suggestion
        self.coords = coords if coords else [0, 0, 0]
```

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo,
+            coords: list[int, int, int] = None
```
pavlejviki, line 57: when you specify a type hint for a list that should contain integers, you don't need to repeat the int type for each element.

pavlejviki (changes requested):

PR #875, opened 2024-02-04
Solution

MaksymNikulyak:


app/main.py
```diff
@@ -1,5 +1,64 @@
...
+
+
 class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] = None) -> None:
```

app/main.py
```diff
@@ -1,5 +1,64 @@
...
+        if coords is None:
```
vasylhnatiuk, line 11: try do it in one line

vasylhnatiuk (changes requested):

app/main.py
MaksymNikulyak: Done - please, check

app/main.py
MaksymNikulyak: Done, thanks!

vasylhnatiuk (approved):

PR #877, opened 2024-02-05
solution

KovbasinskyV:


app/main.py
```diff
@@ -1,5 +1,71 @@
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None,
```
EdAlekseiev, line 14: If argument can be `None` you must show it in type hint. Fix everywhere.
```suggestion
            coords: list | None = None,
```

EdAlekseiev, line 14: Specify inner type of list. Fix everywhere
```suggestion
            coords: list[int] = None,
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #878, opened 2024-02-05
Solution

TarasFirst:


app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
+
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        super().__init__(name=name, weight=weight, coords=coords)
```
Dimosphen1, line 46: You could avoid using multi-line if condition with using `or` expression 
```suggestion
        super().__init__(name=name, weight=weight, coords=coords or [0, 0, 0])
```

app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            coords = [0, 0]
+        self.coords = coords
```
Dimosphen1, line 18: Rewrite this part in one line to avoid using multi-line if condition.

app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot:
+
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
```
Dimosphen1, line 12: Indicate that you expect to return a `list` of `int` in the type annotation.

Dimosphen1 (changes requested):

app/main.py
TarasFirst: Thank you, i fixed

PR #882, opened 2024-02-07
Solution

sudakolha:


app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+
+    def __init__(self, name: str, weight: float, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
```
vasylhnatiuk, line 11: try do it in one line

app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self, name: str, weight: float, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
```
vasylhnatiuk, line 40: same here

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+
+    def __init__(self, name: str, weight: float, coords: list = None) -> None:
```
vasylhnatiuk, line 8:
```suggestion
    def __init__(self, name: str, weight: float, coords: list[int] = None) -> None:
```

vasylhnatiuk: change everywhere

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #883, opened 2024-02-07
Delivery drone: solution

ghost:


app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: str = None) -> None:
```
EdAlekseiev, line 7: If argument can be `None` you should show it in type hint. Fix everywhere.
```suggestion
    def __init__(self, name: str, weight: int, coords: str | None = None) -> None:
```

EdAlekseiev, line 7: Invalid type hint for `coords`. `coords` is a list
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] = None) -> None:
```

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #889, opened 2024-02-09
solution

vanko6226:


app/main.py
```diff
@@ -2,4 +2,56 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
EdAlekseiev, line 7: If argument can be None, you must show it in type hint. Also, specify inner type of list. Fix everywhere.
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] | None = None) -> None:
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #893, opened 2024-02-12
Implement py-delivery-drone

anrozhna:


app/main.py
```diff
@@ -1,5 +1,89 @@
...
+            self,
+            name: str,
+            weight: int,
+            coords: List[int] = None
+    ) -> None:
+        super().__init__(weight)
+        self.name = name
+        self.coords = coords
+        if self.coords is None:
+            self.coords = [0, 0]
```
pavlejviki, line 23: Use ternary operator here

app/main.py
```diff
@@ -1,5 +1,89 @@
...
+
+    def go_forward(
+            self,
+            step: int = 1
+    ) -> None:
+        for i in range(step):
+            self.coords[1] += 1
```
pavlejviki, line 30: there is no need to use a loop, you can deduct total steps straight away not one by one 

fix in all  methods

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,5 +1,81 @@
...
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: List[int] = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+
+        if self.coords is None or len(self.coords) == 2:
+            self.coords = [0, 0, 0]
```
Dimosphen1, line 53: There is no need to redefine the `coords` after initialization in the parent class, pass to the `super().__init__` the right value from the beginning.
```suggestion
        super().__init__(name, weight, coords or [0, 0, 0])
```

app/main.py
```diff
@@ -1,5 +1,81 @@
...
+class BaseRobot(Cargo):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: List[int] = None
+    ) -> None:
+        super().__init__(weight)
+        self.name = name
+        self.coords = coords if coords is not None else [0, 0]
```
Dimosphen1, line 21: You could omit the check `is not None` thus simplifying your code
```suggestion
        self.coords = coords if coords else [0, 0]
```
Or use the `or expression` for the same purpose
```suggestion
        self.coords = coords or [0, 0]
```

Dimosphen1 (changes requested):

PR #897, opened 2024-02-15
Solution

d-kaliuta:


app/main.py
```diff
@@ -2,4 +2,83 @@ class Cargo:
...
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        self.name = name
+        self. weight = weight
+
+        if coords is None:
```
vasylhnatiuk, line 16: do it in oneline

vasylhnatiuk (changes requested):

app/main.py
vasylhnatiuk: do it

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):
read it https://stackoverflow.com/questions/2802726/putting-a-simple-if-then-else-statement-on-one-line

vasylhnatiuk (approved):

PR #900, opened 2024-02-15
Develop

ggeorgiyy:


EdAlekseiev (changes requested):
Fix flake8

app/main.py
```diff
@@ -1,5 +1,64 @@
...
+
+
 class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot():
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
YukihiroSM, line 10: It's better to specify more detailed type annotation
```suggestion
    def __init__(self, name: str, weight: int, coords: Optional[list[int]] = None) -> None:
```

Also, please, note, if default value for the parameter is None - then this parameter should be marked as Optional. Please, fix it everywhere.

app/main.py
```diff
@@ -1,5 +1,64 @@
...
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
YukihiroSM, line 14: It can be transformed into ternary operator:
```suggestion
        self.coords = [0, 0] if coords is None else coords
```
OR:
```suggestion
        self.coords = coords if coords else [0, 0]
```

Fix it in other cases as well, please.

app/main.py
```diff
@@ -1,5 +1,64 @@
...
+
+    def go_down(self, step: int = 1) -> list:
+        self.coords[2] = self.coords[2] - step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self, name: str, weight: int,
+            coords: list = None, max_load_weight: int = None,
+            current_load: int = None) -> None:
```
YukihiroSM, line 53: According to the hook_load logic -> current_load is Cargo, not int.

YukihiroSM (changes requested):

app/main.py
ggeorgiyy: Thanks! Прикольно, сейчас почитаю про Optional...

EdAlekseiev (approved):
Good job

PR #901, opened 2024-02-15
Solution

AlexTsikhun:


app/main.py
```diff
@@ -2,4 +2,71 @@ class Cargo:
...
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int | float,
+            coords: list = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        if not coords:
+            self.coords = [0, 0, 0]
```
Dimosphen1, line 45: Do not redefine `coords` after passing them to the `__init__`, you could evaluate `coords` at the stage of passing them to the `__init__` using `or` expression
```suggestion
        super().__init__(name, weight, coords or [0, 0, 0])
```

app/main.py
```diff
@@ -2,4 +2,71 @@ class Cargo:
...
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int | float,
+            coords: list = None
+    ) -> None:
+        if not coords:
+            coords = [0, 0]
```
Dimosphen1, line 14: Perform assignment of `coords` directly to `self.coords` in one line using `if` condition or `or` expression

Dimosphen1 (changes requested):

app/main.py
AlexTsikhun: > Perform assignment of `coords` directly to `self.coords` in one line using `if` condition or `or` expression

I don't often use `or` expression, but I'll update my code with it! 
But all the time,  I prefered using `if` for redability (imho)

AlexTsikhun: I just have 1 problem when i'm on task - I blindly follow instructions (sometimes without fully understanding of the task (or business logic))

Polyakiv-Andrey (approved):
Good job!

PR #904, opened 2024-02-17
'Solution'

sdannyk:


app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
EdAlekseiev, line 7: Specify inner type of list. Also, if argument can be `None` you must show it in type hint. Fix everywhere.
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] | None = None) -> None:
```

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self, name: str, weight: int, coords: list[int] | None = None
+    ) -> None:
+        if coords is None:
```
vasylhnatiuk, line 34: write it in one line as well

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):

EdAlekseiev (approved):
Good job

PR #907, opened 2024-02-21
Solution

twonumberfortyfives:


app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, creature: Cargo) -> None:
+        if self.current_load is None and \
+                creature.weight <= self.max_load_weight:
+            self.current_load = creature
+        else:
+            self.coords = [0, 0, 0]
```
vsmutok, line 71:
```suggestion
    def hook_load(self, cargo: Cargo) -> None:
        if self.current_load is None and cargo.weight <= self.max_load_weight:
            self.current_load = cargo
```

vsmutok (changes requested):

app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: list = None
+    ) -> None:
+        super().__init__(
+            name,
+            weight,
+            coords=coords or [0, 0, 0]
```
Dimosphen1, line 43: It is a better to use key for all arguments here to not depend on ordering and since you already used `coords` key

Dimosphen1 (approved):

PR #908, opened 2024-02-21
'Solution'

hrynivya:


app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
vsmutok, line 10: use `or` in this case instead of `if ... else`

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
+            self.current_load = cargo
+        else:
+            self.coords = [0, 0, 0]
```
vsmutok, line 59:
```suggestion
    def hook_load(self, cargo: Cargo) -> None:
        if self.current_load is None and cargo.weight <= self.max_load_weight:
            self.current_load = cargo
```

vsmutok: `else` redundant here

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #909, opened 2024-02-21
Drones was created!

itpsychologist:


app/main.py
```diff
@@ -2,4 +2,77 @@ class Cargo:
...
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords=coords[:1])
+        self.coords = coords
```
Dimosphen1, line 43: There is no need to complicate the logic and redefine `coords` after initialization. Use `or` expression instead
```suggestion
        super().__init__(name, weight, coords=coords or [0, 0, 0])
```

app/main.py
```diff
@@ -2,4 +2,77 @@ class Cargo:
...
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: int = None,
+            coords: list = None
+    ) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
```
Dimosphen1, line 62: You could delete this code, setting `coords` is a responsibility of a parent class

app/main.py
```diff
@@ -2,4 +2,77 @@ class Cargo:
...
+            self.current_load = cargo
+
+    def unhook_load(self) -> None:
+        self.current_load = None
+
+
+robot_1 = BaseRobot("", 1)
+robot_2 = BaseRobot("", 1)
+robot_1.go_forward(1)
+robot_2.coords == [0, 0]
```
Dimosphen1, line 78: Remove this code

app/main.py
```diff
@@ -2,4 +2,77 @@ class Cargo:
...
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
Dimosphen1, line 15: You could remove `is not None` here
```suggestion
        self.coords = coords if coords else [0, 0]
```

Dimosphen1 (changes requested):

PR #914, opened 2024-02-23
Solution

Voronessko:


app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
EdAlekseiev, line 7: If argument can be None, you should show it in type hint. Also, specify inner type of list. Fix everywhere
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] | None = None) -> None:
```

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None,
+                 max_load_weight: int = None,
+                 current_load: Cargo = None) -> None:
```
EdAlekseiev, line 53: If argument can be None, you should show it in type hint. Also, specify inner type of list.

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
EdAlekseiev, line 35: If argument can be None, you should show it in type hint. Also, specify inner type of list.

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] | None = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+        if not self.coords:
```
vasylhnatiuk, line 14: write it in one line

vasylhnatiuk (changes requested):

EdAlekseiev (approved):
Good job

PR #919, opened 2024-02-26
Task completed✅

Githi54:
🤖Implemet the BaseRobot
🤖Implemet the FlyingRobot
🤖Implemet the DeliveryDrone

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,5 +1,87 @@
+from typing import List
+
+
 class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: List[int] = None,
+    ) -> None:
+        if coords is None:
+            coords = [0, 0]
+        self.name = name
+        self.weight = weight
+        self.coords = coords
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,5 +1,87 @@
...
+
+class BaseRobot:
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: List[int] = None,
+    ) -> None:
+        if coords is None:
+            coords = [0, 0]
```
LiudmylaKulinchenko, line 17: You can make it in one line, same in every class. Since condition is pretty simple, it's ok to use ternary operator here

app/main.py
```diff
@@ -1,5 +1,87 @@
...
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+        # not unpacking, because while take FlyingRobot coords have a error
+        self.x = coords[0]
+        self.y = coords[1]
```
LiudmylaKulinchenko, line 23: Consider to remove self.x, y and z attributes, looks like it doesn't simplify your code in this task

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Nice! 🌼

PR #920, opened 2024-02-26
implement clases BaseRobot/FlyingRobot/DeliveryDrone

MaxSEngineer:


app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot:
+    def_p_step = 1
+
+    def __init__(self, name: str,
+                 weight: int,
+                 coords: None | list = None
+                 ) -> None:
+        self.coords = coords if coords else [0, 0]
```
vsmutok, line 13: better to use `or` in this case

app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def_p_step = 1
```
vsmutok, line 7: why you need it?
Also `def_p_step` not clear name

vsmutok (changes requested):

app/main.py
MaxSEngineer: Thanks, its really clear way

app/main.py
MaxSEngineer: I thought that i can control value of default increment in methods 
I deleted this

PR #921, opened 2024-02-26
Solution

Viesich:


app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
```
Dimosphen1, line 10: Place each argument including `self` on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability. Apply this change to all classes

app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        coords = coords or [0, 0, 0]
+        if len(coords) < 3:
+            coords += [0] * (3 - len(coords))
+        super().__init__(name, weight, coords)
```
Dimosphen1, line 40: Simplify the logic by defining the value at the time of passing it to `super().__init__`
```suggestion
        super().__init__(name, weight, coords or [0, 0, 0])
```

Dimosphen1 (changes requested):

vasylhnatiuk (approved):

PR #941, opened 2024-03-07
add solution

Dr-Dron:


app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Dimosphen1, line 7: Place each argument including `self` on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability. Apply this change in all classes

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
Dimosphen1, line 10: There is no need to perform the `is not None` check here
```suggestion
        self.coords = coords if coords else [0, 0]
```
Or you can make the assignment even shorter using `or`
```suggestion
        self.coords = coords or [0, 0]
```

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        adjusted_coords = coords if coords is not None else [0, 0, 0]
+        super().__init__(name, weight, adjusted_coords)
```
Dimosphen1, line 31: There is no need to create `adjusted_coords` variable, use or instead
```suggestion
        super().__init__(name, weight, coords or [0, 0, 0])
```

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+            weight: int,
+            coords: list = None,
+            max_load_weight: int = 0,
+            current_load: int = None
+    ) -> None:
+
+        super().__init__(
+            name,
+            weight,
+            coords if coords is not None else [0, 0, 0]
```
Dimosphen1, line 53: You do not need to define which coords should pass further to the parent class this work is already done in the `FlyingRobot` `__init__` method. Just pass `coords` as is to the parent class

Dimosphen1 (changes requested):

PR #944, opened 2024-03-11
Solution

Bujhm99:


app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+        self.coords[2] += step
+
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self, name: str, weight: int,
+                 max_load_weight: int, current_load: Cargo = None,
+                 coords: list = None, ) -> None:
```
Boyok1337, line 49: create a new line for each atribute

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
Boyok1337, line 13: make it to one line with operator "not"
if not coords...
```suggestion
            self.coords = coords[0, 0] if not coords else coords
```

Boyok1337 (changes requested):

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = [0, 0] if not coords else coords
```
YuriiKindrat, line 10: u can write here like this 
`coords or [0,0]`

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
...
+    def get_info(self) -> None:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        else:
+            coords = coords
```
YuriiKindrat, line 33: same is here, also it can be done inside super method

YuriiKindrat (changes requested):

YuriiKindrat (approved):
LGTM!

PR #948, opened 2024-03-12
Solution

Boyok1337:


nicksetrakov (approved):

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = [0, 0] if not coords else coords
```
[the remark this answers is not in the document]
YuriiKindrat: yeah, u can use 'or' to write less code, but it is not necessary

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        super().__init__(name, weight)
+        self.coords = [0, 0, 0] if not coords else coords
```
YuriiKindrat, line 41: no need to rewrite attr here, u can make check inside super method)

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: int,
```
YuriiKindrat, line 56: it seems to me, that current load is not int )

YuriiKindrat (changes requested):

Sofbog (approved):
LGTM

PR #949, opened 2024-03-12
solution

xtsvirko:


app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
```suggestion
        self.coords = coords or [0, 0]
```

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] = None) -> None:
```

app/main.py
vasylhnatiuk: change everywhere

app/main.py
vasylhnatiuk: change everywhere

vasylhnatiuk (changes requested):

app/main.py
AlexanderTsvirko: done

app/main.py
AlexanderTsvirko: done

vasylhnatiuk (approved):

PR #955, opened 2024-03-16
Solution

Vova-Stoliar:


app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] | None = None) -> None:
```
LiudmylaKulinchenko, line 10: You should add a None type annotation only if you want to pass a None as coords here. If you just want to make coords an optional parameter, you should remove None from a type annotation and leave just a default value
Same in other methods

app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
+        self.weight = weight
+        self.name = name
+
+        self.coords = [0, 0] if coords is None else coords
```
LiudmylaKulinchenko, line 14: You can use `if not coords` here, so if you get an empty list, it will still work correctly. Same in other init methods

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Looks great! ✨

PR #963, opened 2024-03-21
Solution

DmytroHlazyrin:


app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        self.name = name
+        self.weight = weight
+        if not coords:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
panicua, line 16: I don't know if you should change it, but.. it will look better if you do it in 1 line ```self.coords = ... if not ... esle ...```

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        super().__init__(name, weight, coords)
+        if not coords:
+            self.coords = [0, 0, 0]
+        else:
+            self.coords = coords
```
panicua, line 43: same here as above if you will want to change it.

panicua (approved):
I'm not a pro, so I asked about those approaches chat GPT, idk if it answered me right but..
```In general, both approaches achieve the same result, so the choice depends on which style you and your team find more readable and maintainable.```

File: app/main.py (hunk 1 above)
Comment (YuriiKindrat) [reply app/main.py:16]: yeah, @panicua is right
u could do this in 1 line 
self.coords = coords or [0,0]

File: app/main.py (hunk 2 above)
Comment (YuriiKindrat) [inline app/main.py:43]: same, and inside the super method

**Review by YuriiKindrat — CHANGES_REQUESTED** [review]

File: app/main.py [hunk 3]
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = [0, 0, 0] if not coords else coords
```
LiudmylaKulinchenko, line 37: You can do it like this ` coords or [0,0]`, like Yurii suggested and place directly inside super

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):
GJ!

PR #964, opened 2024-03-21
Solution!

scandriy:


app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```

app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
+        else:
+            self.coords = coords
```
EdAlekseiev, line 37: The same comment as above

app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
...
+            name: str,
+            weight: int,
+            coords: list = None,
+            max_load_weight: int = None,
+            current_load: Cargo = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+        super().__init__(name, weight, coords)
```
EdAlekseiev, line 58: Don't call `__init__` twice

EdAlekseiev (changes requested):

app/main.py
scandriy: I'm not sure what's you mean, beacuse with '==' operator it doesn't pass flake8 check and in other case 'not' will be probably the same as 'is', so please explaine in details

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        if coords == None:
+            self.coords = [0, 0, 0]
+        else:
+            self.coords = coords
```
EdAlekseiev, line 37: Rhe same comment as above

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords == None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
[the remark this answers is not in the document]
scandriy: So, I guess it all about in one line option or it make any logical sens?

EdAlekseiev (approved):

PR #965, opened 2024-03-21
Solution

panicua:


Gilgumish (approved):
GJ!

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: float | int,
+                 coords: list = None) -> None:
+        super().__init__(name=name, weight=weight, coords=coords)
+        self.coords = [0, 0, 0] if not coords else coords
```
YuriiKindrat, line 37: no need to overwrite attr here, u could make a check inside super method

YuriiKindrat (changes requested):

app/main.py
panicua: > no need to overwrite attr here, u could make a check inside super method

Done, but won't it affect the readability?

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: float | int,
+                 coords: list = None) -> None:
+        super().__init__(name=name,
+                         weight=weight,
+                         coords=[0, 0, 0] if not coords else coords)
```
LiudmylaKulinchenko, line 38: You can use `coords or [0, 0, 0]` here to improve readability

LiudmylaKulinchenko (approved):
Well done 🌸

PR #970, opened 2024-03-23
Solution

y-kondrashova:


Chub-andrew (approved):
Looks good.

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
YuriiKindrat, line 13: this could be done in one line 
self.coords  = coords or [0,0]

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
+        else:
+            self.coords = coords
```
YuriiKindrat, line 37: same is here, also there is no need to overwrite attr, u could make a check directly inside thie super method

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: object | None,
```
YuriiKindrat, line 51: everything in python is object, what did u mean ?)

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
...
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = coords or [0, 0, 0]
```
LiudmylaKulinchenko, line 31: You missed put it directly inside super method instead of overwriting the attribute

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -2,4 +2,54 @@ class Cargo:
...
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int | None,
```
YuriiKindrat, line 44: it seems to me, that current load should be Cargo type)

YuriiKindrat (changes requested):

PR #971, opened 2024-03-23
Solution

IlliaPetukhov:


app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] = None) -> None:
```

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
...
+        if coords is None:
```
vasylhnatiuk, line 8: do it in one line

app/main.py
vasylhnatiuk: do it everywhere

vasylhnatiuk (changes requested):

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #976, opened 2024-03-24
Solution#1

VitalyBashkiser:


siaserb (approved):

app/main.py
```diff
@@ -1,5 +1,69 @@
...
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        else:
+            if len(coords) < 3:
+                coords.extend([0] * (3 - len(coords)))
```
YuriiKindrat, line 38: this could be done inside the super method 
hint: coords or [0,0,0]

YuriiKindrat (changes requested):

app/main.py
VitalyBashkiser: Thanks for the advice)

YuriiKindrat (approved):
GJ!

PR #977, opened 2024-03-24
Solution

siaserb:


VitalyBashkiser (approved):
At first glance, everything is not bad

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0]
```
YuriiKindrat, line 9: this could be done like this:
self.coords = coords or [0,0]

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        if self.coords is None:
+            self.coords = [0, 0, 0]
+        elif len(self.coords) == 2:
+            self.coords.append(0)
```
YuriiKindrat, line 36: no need to overwrite attr here
u could make a check inside super method

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+        self.coords[2] += step
+
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 max_load_weight: int,
+                 current_load: int,
```
YuriiKindrat, line 48: is current load is int ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ!

PR #978, opened 2024-03-24
solution

alonasorochynska:


semermaks (approved):
GJ

app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
```
Oleksl888, line 10: you can set `self.coorsds in one line and avoid using `if` statements at all (use or comparison to evaluate `coords`)

app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        super().__init__(name, weight, coords)
```
Oleksl888, line 38: same as above - you can simplify this with `or` expression and may will all fit in one line in `super()` call. This way you also wont need to reassign `coords` value twice

app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: object = None,
```
Oleksl888, line 57: what would be the right type for `current_load`?

Oleksl888 (changes requested):

app/main.py
YuriiKindrat: u didin't fix that
try to do smth like that:
`self.coords = coords or [0,0]`

app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = [0, 0, 0] if coords is None else coords
```
YuriiKindrat, line 39: same is here, no need to overwrite attr, u could make a check inside super method

app/main.py
YuriiKindrat: try to avoid using object type annotation - everything in python is an object

YuriiKindrat (changes requested):

YuriiKindrat (approved):
LGTM!

PR #979, opened 2024-03-25
Solution

Nyxan:


Ruxxell54 (approved):
Great

app/main.py
```diff
@@ -1,5 +1,60 @@
...
+        self.coords[0] -= step
+
+    def get_info(self) -> None:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
```
YuriiKindrat, line 34: no need to make check before the super, make this inside super method )

app/main.py
```diff
@@ -1,5 +1,60 @@
...
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int = None,
```
YuriiKindrat, line 49: it seems to me, that this is wrong type annotation for current_load)

YuriiKindrat (changes requested):

app/main.py
YuriiKindrat: hint: coords or [0,0,0]

app/main.py
```diff
@@ -1,5 +1,58 @@
...
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
Oleksl888, line 13: As Yuri mentioned before you may simplify these conditionals with `self.coords = coords or [0, 0]` If coords is None  python will evaluate next expression after or and assign it

app/main.py
```diff
@@ -1,5 +1,58 @@
...
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: Cargo,
+                 coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
```
Oleksl888, line 54: You can also use `if not self.current_load` in cases like these

Oleksl888 (approved):
you passed

app/main.py
Nyxan: Got it, thanks

PR #980, opened 2024-03-25
Solution 1 (add all classes)

semermaks:


app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
...
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.coords = coords if coords is not None else [0, 0, 0]
```
Nyxan, line 30: just check for None, because we have coords in super method
`if coords is None:`
`coords = [0, 0, 0]`

Nyxan (changes requested):

app/main.py
YuriiKindrat: u can make a check directrly in super method, no need to create attr 
hint: coords or [0,0,0]

app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
...
+
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self, name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int,
```
YuriiKindrat, line 44: is current load is int ?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
...
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.coords = coords or [0, 0, 0]
```
YuriiKindrat, line 30: no need to overwrite attr here, u could do this check inside super method

YuriiKindrat (changes requested):

Oleksl888 (approved):
ll done sir!

PR #982, opened 2024-03-25
'Solution'

dkotkod:


app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
+                 weight: int,
+                 coords: list[int] = None,
+                 max_load_weight: int = 0,
+                 current_load: Cargo = None) -> None:
+        super().__init__(name, weight, coords=coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
```
SerhiiMar, line 59: Here you can just write: `not self.current_load`

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = [0, 0] if coords is None else coords
```
SerhiiMar, line 13: You can invert `if` statement and write it more simple

SerhiiMar (approved):
GJ!

app/main.py
YuriiKindrat: or u could just write like this 
`self.coords = coords or [0,0]`

YuriiKindrat (approved):

PR #984, opened 2024-03-25
'Solution'

yossosh:


app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
```
alonasorochynska, line 41: I am not sure but probably super() have to be first after init

alonasorochynska (approved):
check comment please

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
```
alonasorochynska, line 35: I am not sure but probably super() have to be first after init

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+                 name: str,
+                 weight: int,
+                 coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
Oleksl888, line 16: you can set self.coords in one line and avoid using if statements at all (use `or` comparison to evaluate `coords`)

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None) -> None:
+        if coords is None:
```
Oleksl888, line 39: same as above - you can simplify this with `or` expression and may will all fit in one line in super() call

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 coords: list = None,
+                 current_load: int = None) -> None:
```
Oleksl888, line 56: double check what is the type for `current_load`

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
```
Oleksl888, line 62: you can also use if `not self.current_load` expression here

Oleksl888 (changes requested):

YuriiKindrat (approved):
LGTM!

PR #985, opened 2024-03-25
Implement Delivery drone task

bwaah1:


messedxup (changes requested):
You need to fix flake8 errors that occurred during tests

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            coords = [0, 0]
+        self.coords = coords
```
Oleksl888, line 12: you can simplify this in one line with ternary operator or use expression with `or`

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+        return (f"Robot: {self.name}, "
+                f"Weight: {self.weight}")
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight)
+        if coords is None:
+            coords = [0, 0, 0]
+        self.coords = coords[:3]
```
Oleksl888, line 36: same as above - you can set coords in one line. Also, why do you need list slicing here?

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None,
+                 max_load_weight: int = 0,
+                 current_load: Cargo = None) -> None:
```
Oleksl888, line 51: it doesn't specifiy in task description to set `current_load` default value

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
```
Oleksl888, line 57: `if self.current_load` also works here

Oleksl888 (changes requested):

PR #988, opened 2024-03-25
'Solution'

kssshevchenko:


mikhailyemets (approved):

app/main.py
```diff
@@ -1,5 +1,56 @@
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
```
Oleksl888, line 27: you can pass coords to super in one line using ternary operator or using brachless approach with `or`

app/main.py
```diff
@@ -1,5 +1,56 @@
...
+
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self, name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: list = None,
```
Oleksl888, line 40: check correct type for current load

app/main.py
```diff
@@ -1,5 +1,56 @@
...
+                 coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: "Cargo") -> None:
```
Oleksl888, line 46: You won't need to use double quotes here if you write your code below `Cargo` class

Oleksl888 (changes requested):

YuriiKindrat (approved):
GJ!

PR #989, opened 2024-03-25
'Solution'

afaidiuk:


kssshevchenko (approved):

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        if coords is None:
+            coords = [0, 0]
```
Oleksl888, line 12: You can set coords here in one line. Use ternary operator or expression with `or`

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        else:
+            if len(coords) == 2:
+                coords.append(0)
+        super().__init__(name, weight, coords)
```
Oleksl888, line 43: Same as above - simplify this, this logic can fit in one line

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 coords: list[int] = None,
+                 current_load: list[int] = None) -> None:
```
Oleksl888, line 58: What should be correct type for current_load?

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,5 +1,67 @@
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        coords = [0, 0, 0] if coords is None else \
```
Oleksl888, line 40: simplify this condition  - no need for second if-else, also do not use `\` for line breakers

Oleksl888 (changes requested):

PR #991, opened 2024-03-27
Solution

VladyslavKsh:


app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+        if self.coords is None:
+            self.coords = [0, 0]
```
Oleksl888, line 12: You can do one assignment in one line. Use ternary operator or use `or` comparison

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
```
Oleksl888, line 32: same as above

Oleksl888 (changes requested):

app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
...
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        coords = [0, 0, 0] if coords is None else coords
```
YuriiKindrat, line 30: no need to make ckeck here, u could do this inside super method like this 
`coords or [0,0,0]`

YuriiKindrat (changes requested):

app/main.py
VladyslavKsh: Didn't think about it, thanks)

YuriiKindrat (approved):
GJ!

PR #992, opened 2024-03-27
solved

serhii-chaban:


app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
EdAlekseiev, line 7: If argument can be None you should show it in type hint. Also, specify iiner type of list. Fix everywhere
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] | None = None) -> None:
```

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
EdAlekseiev, line 29: The same comment as above

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: int | None,
+            coords: list = None
```
EdAlekseiev, line 47: The same comment as above

EdAlekseiev (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #999, opened 2024-04-03
solution

ArtemLeo:


app/main.py
```diff
@@ -1,5 +1,72 @@
...
+            weight: int,
+            max_load_weight: int,
+            current_load: "Cargo" = None,
+            coords: List[int] = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: "Cargo") -> None:
```
Dimosphen1, line 67: Do not define class inside `""` in the type annotation
```suggestion
    def hook_load(self, cargo: Cargo) -> None:
```
Use `from __future__ import annotations` in case you face errors. Apply similar change to the other places as well

app/main.py
```diff
@@ -1,5 +1,72 @@
...
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: List[int] = None
+    ) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
```
Dimosphen1, line 45: Do not create if condition, you could use `or` instead
```suggestion
        super().__init__(name, weight, coords or [0, 0, 0])
```

app/main.py
```diff
@@ -1,5 +1,72 @@
...
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: List[int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
Dimosphen1, line 18: You could simplify this condition
```suggestion
        self.coords = coords if coords else [0, 0]
```

Dimosphen1 (changes requested):

vsmutok (approved):
Good Job!

PR #1002, opened 2024-04-05
Solution

dvkim28:


app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords.copy() if coords else [0, 0]
```
vasylhnatiuk, line 11:
```suggestion
        self.coords = coords or [0, 0]
```

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
vasylhnatiuk, line 8:
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] = None) -> None:
```

app/main.py
vasylhnatiuk: do everywhere

app/main.py
vasylhnatiuk: do it everywhere

vasylhnatiuk (changes requested):

vsmutok (approved):
Great!

PR #1008, opened 2024-04-08
Solution

sind14:


app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
vsmutok, line 13: better to use `or` in this case

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight)
+        if coords is None:
+            self.coords = [0, 0, 0]
+        else:
+            self.coords = coords
```
vsmutok, line 37: better to use `or` in this case

vsmutok (changes requested):

app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
EdAlekseiev, line 7: If argument can be `None`, you should show it in type hint. Also, you should specify inner type of list. Fix everywhere, please
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] | None = None) -> None:
```

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: Cargo = None,
```
EdAlekseiev, line 51:
```suggestion
                 current_load: Cargo | None = None,
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #1010, opened 2024-04-10
Solution

Lebid-Dmytro:


app/main.py
```diff
@@ -1,5 +1,71 @@
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: Optional[Iterable[int]] = None
+    ) -> None:
+        coords = coords or [0, 0, 0]
+        super().__init__(name, weight, coords)
```
Dimosphen1, line 44: There is no need to create `coords` variable here, make an `or`  evaluation directly in the `super().__init__(`

Dimosphen1 (changes requested):
Several minor changes were requested

PR #1031, opened 2024-05-08
Solution

Shayba228:


app/main.py
```diff
@@ -1,5 +1,61 @@
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
vsmutok, line 16: better to use `or` in this case

app/main.py
```diff
@@ -1,5 +1,61 @@
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
+        else:
+            self.coords = coords
```
vsmutok, line 40: better to use `or` in this case

app/main.py
```diff
@@ -1,5 +1,61 @@
+from __future__ import annotations
```
vsmutok, line 1: I think this import redundant here

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #1032, opened 2024-05-08
Solution

TsvetanKichuk:


app/main.py
```diff
@@ -1,5 +1,71 @@
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
EdAlekseiev, line 2: You should specify inner type of list. If argument can be None, you should show it in type hint. Fix everywhere
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] | None = None) -> None:
```

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
EdAlekseiev, line 8: You can simplify your code
```suggestion
        self.coords = coords or [0, 0]
```

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,5 +1,80 @@
...
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] | None = None
+    ) -> None:
+
+        if coords is None:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
```
Dimosphen1, line 39: There is no need to add an if condition here use `or` here as well
```suggestion
        super().__init__(name, weight, coords or [0, 0, 0])
```

Dimosphen1 (changes requested):

vasylhnatiuk (approved):

PR #1033, opened 2024-05-11
hell yeah solution

RuslanLevchuk:


app/main.py
```diff
@@ -1,5 +1,74 @@
...
+            self, name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        if coords:
+            self.coords = coords
+        else:
+            self.coords = [0, 0]
```
vsmutok, line 20: better to use `or` in this case

app/main.py
```diff
@@ -1,5 +1,74 @@
...
+    def __init__(
+            self, name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        super().__init__(name=name, weight=weight)
+        if coords:
+            self.coords = coords
+        else:
+            self.coords = [0, 0, 0]
```
vsmutok, line 48: better to use `or` in this case

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1035, opened 2024-05-15
'Solution'

AlexTarasov57:


app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = [0, 0] if coords is None else coords
```
Dimosphen1, line 15: You could write this condition without `is`
```suggestion
        self.coords = coords if coords else [0, 0]
```
Or using `or`
```suggestion
        self.coords = coords or [0, 0]
```

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        super().__init__(name=name, weight=weight)
+        self.coords = [0, 0, 0] if coords is None else coords
```
Dimosphen1, line 41: There is no need to redefine self.coords after intialization, perform assignment in the `super().__init(`
```suggestion
        super().__init__(name=name, weight=weight, coords=coords or [0, 0, 0])
```

Dimosphen1 (changes requested):

PR #1038, opened 2024-05-19
added solution to delivery drone

AlexRozv:


app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            coords = [0, 0]
+        self.coords = coords
```
LiudmylaKulinchenko, line 12: Use [ternary operator](https://www.geeksforgeeks.org/ternary-operator-in-python/) here and in the next init methods. It will be readable and clear in this case

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int,
```
LiudmylaKulinchenko, line 48: Is current_load integer?

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = [0, 0] if coords is None else coords
```
LiudmylaKulinchenko, line 10: What if coords is blank list? Let's include it in the condition (`if coords` instead of `if coords is None`)

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job ✨

PR #1040, opened 2024-05-22
solution

AndrewwSSS:


app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str,
+                 weight: int,
+                 coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords:
+            self.coords = coords
```
DevSpaciX, line 13: Try to write it in one line )

DevSpaciX (changes requested):

app/main.py
AndrewwSSS: done.

DevSpaciX (approved):
Good job!

PR #1043, opened 2024-05-25
Solution

Jejemira:


app/main.py
```diff
@@ -2,4 +2,106 @@ class Cargo:
...
+            name: str,
+            weight: int,
+            coords: list = [0, 0]
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+
+    def go_forward(self, number: int = 1) -> None:
+        if len(self.coords) == 3:
```
vasylhnatiuk, line 18: DRY, don’t repeat yourself, you can avoid this check in every method

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -2,4 +2,100 @@ class Cargo:
...
+            name: str,
+            weight: int,
+            coords: list = [0, 0]
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+
+    def go_forward(self, number: int = 1) -> None:
+        self.coords = [
```
vasylhnatiuk, line 18: self.coords[1] += number

app/main.py
```diff
@@ -2,4 +2,100 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = [0, 0]
```
vasylhnatiuk, line 11: It is not a good practice to use mutable object as a default parameter, set it to None by default, and use condition

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -8,43 +8,25 @@ def __init__(
             self,
             name: str,
             weight: int,
-            coords: list = [0, 0]
+            coords: list = None
     ) -> None:
         self.name = name
         self.weight = weight
+        if coords is None:
```
vasylhnatiuk, line 15: do it in one line, use `or`

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1046, opened 2024-05-26
Solution

necobra:


app/main.py
```diff
@@ -2,4 +2,54 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0]
+        self.name = name
+        self.weight = weight
+        self.coords = coords
```
Oleg-s-s, line 12: You can write coords in 1 line, read about if else one liners

Oleg-s-s (changes requested):

Oleg-s-s (changes requested):
.

app/main.py
```diff
@@ -2,4 +2,53 @@ class Cargo:
...
+
+    def go_right(self, value: int = 1) -> None:
+        self.coords[0] += value
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
DevSpaciX, line 30: 1.Also think what type coords can be besides list

app/main.py
```diff
@@ -2,4 +2,53 @@ class Cargo:
...
+        if coords is None:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
```
DevSpaciX, line 33: Same with `coords` here ) You can write in default param your check but using operator `or` or one liners as you did on line 8 : )

DevSpaciX (changes requested):

app/main.py
necobra: 1 maybe dict also

DevSpaciX (changes requested):

app/main.py
DevSpaciX: I mean it could be None -- > `coords: list[int] | None = None` will be better , no ?

app/main.py
```diff
@@ -2,4 +2,54 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int,
+                 coords: list[int] | None = None) -> None:
+        coords = coords or [0, 0]
```
Oleg-s-s, line 9: or like this, you have a point here))

Oleg-s-s (approved):

PR #1047, opened 2024-05-26
Solution

Alex93Troy:


app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str,
+                 weight: int,
+                 coords: list = None
+                 ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
vasylhnatiuk, line 13: Write it ine obe l
```suggestion
        self.coords = coords or [0, 0]
```

vasylhnatiuk: Do it everywhere

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str,
+                 weight: int,
+                 coords: list = None
```
vasylhnatiuk, line 9:
```suggestion
                 coords: list[int] = None
```

vasylhnatiuk (changes requested):

app/main.py
Alex93Troy: 👌  okay

app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str,
+                 weight: int,
+                 coords: list[int] = None
+                 ) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
```
vsmutok, line 38: better to use `or` in this case

vsmutok (changes requested):

vsmutok (approved):
Good Job!

PR #1048, opened 2024-05-27
Solution

MaksymProtsak:


app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
```
vasylhnatiuk, line 10: Try to do it in one line

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] = None) -> None:
```

vasylhnatiuk (changes requested):

app/main.py
MaksymProtsak: `self.coords = coords if coords else [0, 0]`

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords else [0, 0]
```
vasylhnatiuk, line 13:
```suggestion
        self.coords = coords or [0, 0]
```

vasylhnatiuk: it is better to use or in this case

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
```
vasylhnatiuk, line 34: do the same here

vasylhnatiuk (changes requested):

app/main.py
vasylhnatiuk: fix it

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1050, opened 2024-05-27
'Solution'

BohdanKuzik:


app/main.py
```diff
@@ -1,5 +1,62 @@
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
```
DevSpaciX, line 5: You can simplify your code using operator `or`

app/main.py
```diff
@@ -1,5 +1,62 @@
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
DevSpaciX, line 27: 1. Same with `coords` here ) You can write in default param your check on line 28-29 but using operator `or`

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,5 +1,62 @@
...
+        if coords is None:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
```
DevSpaciX, line 30: Same with `coords` here ) You can write in default param your check on line 28-29 but using operator `or`

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,5 +1,68 @@
...
         self.weight = weight
 
-# write your code here
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None,
```
DevSpaciX, line 55: `coords: list[int] | None = None` will be better , no ?

DevSpaciX (approved):
Goog job! But read the comment pls

PR #1051, opened 2024-05-27
Solution

Yuruch:


app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
...
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weigh: int,
+            coords: list = None
+    ) -> None:
+        self.name = name
+        self.weight = weigh
+        if coords is None:
```
DevSpaciX, line 15: You can write in in one line using operator `or`

app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
```
DevSpaciX, line 41: 1.Also think what type coords can be besides list

app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
...
+    ) -> None:
+        super().__init__(name, weight, coords)
+        if len(self.coords) != 3:
+            self.coords.append(0)
```
DevSpaciX, line 45: You can write it `super().__init__(name=name, weight=weight, coords=coords or [0, 0, 0]) `Do not you ?

DevSpaciX (changes requested):

app/main.py
Yuruch: image
так чомусь не працює, з позиційними все ок

DevSpaciX: Cuz your param named "weigh" : )))

Yuruch: i see) thanks

DevSpaciX (approved):
Perfect !

PR #1052, opened 2024-05-27
solution

VShveda:


app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
DevSpaciX, line 13: You can write it in one line using operator `or`

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        if len(self.coords) == 2:
+            self.coords.append(0)
```
DevSpaciX, line 35: You can write it `super().__init__(name=name, weight=weight, coords=coords or [0, 0, 0]) `Do not you ?

DevSpaciX (changes requested):

app/main.py
VShveda: I will do it

DevSpaciX (changes requested):
A little type hints optimization : )

app/main.py
```diff
@@ -2,4 +2,55 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int = None,
+                 coords: list = None
```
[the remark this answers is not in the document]
VShveda: it is written in the task description "It is not a good practice to use mutable object as a default parameter".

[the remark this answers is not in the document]
DevSpaciX: I mean `coords: list[int] | None = None`

DevSpaciX (approved):
Good job , read the comment please : )

PR #1053, opened 2024-05-27
Solution_V1

BileichukIvan:


app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
DevSpaciX, line 10: It can be `coords` or `[0, 0]`

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
DevSpaciX, line 29: 1. Same with `coords` here ) You can write in default param your check on line 30-31 but using operator `or`

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
...
+        if coords is None:
+            coords = [0, 0, 0]
+        elif len(coords) == 2:
```
DevSpaciX, line 32: Is this check is necessary ? What if you remove it ?)

DevSpaciX (changes requested):

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
...
+            coords.append(0)
+        super().__init__(name, weight, coords)
```
DevSpaciX, line 34: Same with `coords` here ) You can write in default param your check on line 30-31 but using operator `or`

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None
+                 ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = (coords or [0, 0])
```
DevSpaciX: Could you remove this brackets ?)

DevSpaciX (approved):

PR #1054, opened 2024-05-27
add solution

jpdreamthug:


app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0]
+        self.name = name
+        self.weight = weight
+        self.coords = coords
```
DevSpaciX, line 15: You can simplify your code using operator `or`

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None) -> None:
```
DevSpaciX, line 38: 1. Same with `coords` here ) You can write in default param your check on line 40-41 but using operator `or`

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
+            self.current_load = cargo
+
+    def unhook_load(self) -> None:
+        if self.current_load:
```
DevSpaciX, line 68: That check is redundat imho : )

DevSpaciX (changes requested):

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
```
DevSpaciX, line 41: Same with `coords` here ) You can write in default param your check on line 40-41 but using operator `or`

PR #1055, opened 2024-05-27
Solution

lilarin:


app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
```
DevSpaciX, line 38: 1.Also think what type coords can be besides list

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+    ) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
```
DevSpaciX, line 42: Same with `coords` here ) You can write in default param your check but using operator `or`

DevSpaciX (changes requested):

PR #1056, opened 2024-05-27
Solution

AntonKorinchuk:


app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
...
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        self.name = name
+        self.weight = weight
+        if not coords:
+            coords = [0, 0]
+        self.coords = coords
```
DevSpaciX, line 15: You can simplify your code using operator `or`

app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
```
DevSpaciX, line 37: 1. Same with `coords` here ) You can write in default param your check on line 38-39 but using operator `or`

DevSpaciX (changes requested):

app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
...
+        if not coords:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
```
DevSpaciX, line 40: Same with `coords` here ) You can write in default param your check on line 38-39 but using operator `or`

app/main.py
AntonKorinchuk: The condition states that it will be a list, but we can also use a dictionary, but the code will need to be changed

DevSpaciX (approved):

PR #1057, opened 2024-05-27
Solution

YevaYe:


app/main.py
```diff
@@ -2,4 +2,51 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
DevSpaciX, line 10: You can simplify your code using operator `or`

app/main.py
```diff
@@ -2,4 +2,51 @@ class Cargo:
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
DevSpaciX, line 29: 1. Same with `coords` here ) You can write in default param your check on line 30-31 but using operator `or`
3.List of what data type can be ?

DevSpaciX (changes requested):

app/main.py
```diff
@@ -2,4 +2,51 @@ class Cargo:
...
+        super().__init__(name, weight,
+                         coords if coords is not None else [0, 0, 0])
```
DevSpaciX, line 31: Same with `coords` here ) You can write in default param your check but using operator `or`

app/main.py
```diff
@@ -2,4 +2,52 @@ class Cargo:
...
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int,
+                 coords: list | tuple = None) -> None:
```
DevSpaciX, line 31: It can be `None` instead of `tuple` ) If Your default param `None` , it can be `None` right ?)

DevSpaciX (approved):

PR #1058, opened 2024-05-27
'Solution'

Irina17191:


app/main.py
```diff
@@ -1,5 +1,66 @@
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str,
+                 weight: int,
+                 coords: list[int] | None = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
```
DevSpaciX, line 42: Same with `coords` here ) You can write in default param your check but using operator `or`

DevSpaciX (changes requested):

app/main.py
Irina17191: I got, thank you!)

PR #1064, opened 2024-05-28
'Solution'

Oleh323:


DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,5 +1,64 @@
...
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self, name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: Cargo,
+                 coords: list = [0, 0, 0]) -> None:
```
DevSpaciX, line 54: It was good I mean `coords: list[int] | None = None` will be better , no ?

DevSpaciX (approved):

PR #1065, opened 2024-05-28
delivery-drone solution

DimaYankovskiy:


app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        if not coords:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
```
DevSpaciX, line 42: You can put coords right in `super` method using operator `or`

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo,
+            coords: list = None
```
DevSpaciX, line 58: `coords: list[int] | None = None` will be better , no ?

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #1067, opened 2024-05-28
Solution

NikitaNikolaitsev:


app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self, name: str, weight: int, coords: list = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords else [0, 0]
```
DevSpaciX, line 12: You can use operator `or` and it will be more cleaner imho

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: int,
+            coords: list = None
```
DevSpaciX, line 50: 1.Also think what type coords can be besides list

DevSpaciX (changes requested):

app/main.py
NikitaNikolaitsev: I have no ideas for implementing

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self, name: str, weight: int, coords: list = None  # IDK
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords or [0, 0]  # you mean this?
```
DevSpaciX, line 12: Yeah , I mean this

app/main.py
DevSpaciX: I mean `coords : list[int] | None = None` : )

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords if coords else [0, 0, 0])
```
DevSpaciX, line 33: Why you did not  use `or` like on line 12 ?

DevSpaciX (changes requested):

DevSpaciX (approved):
Great !

PR #1070, opened 2024-05-28
Done task!

TkachukTara:


app/main.py
```diff
@@ -3,3 +3,58 @@ def __init__(self, weight: int) -> None:
...
+        self.coords[2] += step
+
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self, name: str, weight: int,
+                 max_load_weight: int,
+                 current_load: int, coords: list = None) -> None:
```
DevSpaciX, line 48: `coords: list[int] | None = None` will be better , no ?

app/main.py
```diff
@@ -3,3 +3,58 @@ def __init__(self, weight: int) -> None:
...
+        if coords is None:
```
DevSpaciX, line 49: I think you do not need it here , am I wrong ?

app/main.py
```diff
@@ -3,3 +3,58 @@ def __init__(self, weight: int) -> None:
...
 
 # write your code here
+
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            coords = [0, 0]
```
DevSpaciX, line 13: You can write it in one line using operator `or`

DevSpaciX (changes requested):

app/main.py
```diff
@@ -3,3 +3,56 @@ def __init__(self, weight: int) -> None:
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int,
+                 coords: list[int] | None = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
```
DevSpaciX, line 35: Why you did not optimize the code as you did on line 13 ?)

DevSpaciX (changes requested):
Optimize pls that little mistake : )

app/main.py
```diff
@@ -3,3 +3,55 @@ def __init__(self, weight: int) -> None:
...
+class DeliveryDrone(FlyingRobot):
+    def __init__(self, name: str, weight: int,
+                 max_load_weight: int,
+                 current_load: int, coords: list[int] | None = None) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
```
Oleg-s-s, line 53: Your way is correct, but in python we can write shorter
```suggestion
        if not self.current_load and cargo.weight <= self.max_load_weight:
```

PR #1071, opened 2024-05-28
"Solution"

MaglorFeanorson:


app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
+
 # write your code here
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
DevSpaciX, line 11: `self.coords = coords or [0, 0]` will be better , no ?)))

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        elif len(coords) == 2:
+            coords.append(0)
+        super().__init__(name, weight, coords)
```
DevSpaciX, line 35: You can use here operator `or` too , also `elif` block is redundat imho

DevSpaciX (changes requested):

DevSpaciX (approved):

DevSpaciX (approved):

PR #1075, opened 2024-05-30
Solution Delivery drone

lLiftoffl:


app/main.py
```diff
@@ -1,5 +1,68 @@
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None
```
EdAlekseiev, line 5: 1. You should specify inner type of list in type hint
2. If argument can be None, you should show it in type hint
You have to fix it everywhere
```suggestion
                 coords: list[int] | None = None
```

EdAlekseiev (changes requested):

YukihiroSM (approved):
Looks pretty good to me!

PR #1078, opened 2024-06-01
solution

AndreiVed:


app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None
+                 ) -> None:
+        super().__init__(name, weight)
+        self.coords = coords if coords is not None else [0, 0, 0]
```
vsmutok, line 39: better to use `or` in this case

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None
+                 ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
vsmutok, line 14: better to use `or` in this case

vsmutok (changes requested):

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None
```
```suggestion
                 coords: list[int] = None
```

vasylhnatiuk: write everywhere

vasylhnatiuk (changes requested):

vsmutok (approved):
Great!

PR #1084, opened 2024-06-06
Solution

Artemq77:


app/main.py
```diff
@@ -2,4 +2,54 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int,
+                 coords: list[int] = None) -> None:
```
Dimosphen1, line 8: Place each argument including `self` on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability. Apply the same change in the other places as well

app/main.py
```diff
@@ -2,4 +2,54 @@ class Cargo:
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int,
+                 coords: list[int] = None) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = coords or [0, 0, 0]
```
Dimosphen1, line 33: Do not redefine coords, pass the right value from the beginning
```suggestion
        super().__init__(name, weight, coords or [0, 0, 0])
```

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,5 +1,95 @@
...
+        coords: list[int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords or [0, 0]
+
+    def go_forward(
+        self,
+        step: int = 1
+    ) -> None:
```
Dimosphen1, line 23: You could place on a new line only arguments in the `__init__` when there are more than 3 of them. In other cases, you could place them in one line

Dimosphen1 (approved):

PR #1086, opened 2024-06-07
created delivery drone classes

Timur5050:


app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: str,
+            max_load_weight: int,
+            current_load: int,
+            coords: list[int] = None
+    ) -> None:
+        super().__init__(name, weight, coords if coords else [0, 0, 0])
```
Dimosphen1, line 58: There is no need to pass if else condition here, since this logic is already implemented in the parent class 
```suggestion
        super().__init__(name, weight, coords)
```

app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: str,
+            coords: list[int] = None
+    ) -> None:
+        super().__init__(name, weight, coords if coords else [0, 0, 0])
```
Dimosphen1, line 40: You could simplify this line using `or`
```suggestion
        super().__init__(name, weight, coords or [0, 0, 0])
```

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: str,
```
pavlopro, line 37: Correct annotation for weight

app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
+
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: str,
```
pavlopro, line 53: Correct annotation for weight

pavlopro (changes requested):

PR #1087, opened 2024-06-10
Solution

D0mln0:


app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
EdAlekseiev, line 7: You should specify inner type of list. If argument can be None you should show in type hint. Fix everywhere, please
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] | None = None) -> None:
```

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: "Cargo") -> str:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
+            self.current_load = cargo
+            return f"{self.current_load} is {cargo}"
+        else:
+            return f"didn't hook {cargo}"
```
EdAlekseiev, line 65: You don't need else, because if contains return
```suggestion
        if self.current_load is None and cargo.weight <= self.max_load_weight:
            self.current_load = cargo
            return f"{self.current_load} is {cargo}"
        return f"didn't hook {cargo}"
```

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #1091, opened 2024-06-17
+

AndriZhok:


app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None) -> None:
```
Dimosphen1, line 10: Place each argument including `self` on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability. Apply the same change to the other classes as well

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+                 current_load: int = 0) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> bool:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
+            self.current_load = cargo
+            return True
+        return False
```
Dimosphen1, line 66: There is no need to return boolean values here, remove this logic to return `None` instead as it is done in the `unhook_load` method

Dimosphen1 (changes requested):

app/main.py
AndriZhok: line too long (81 > 79 characters)

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+        self.coords[2] += step
+        return self.coords
+
+    def go_down(self, step: int = 1) -> list:
+        self.coords[2] -= step
+        return self.coords
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
```
Polyakiv-Andrey, line 46: After opening the parentheses, write each the arguments on a new line, including the first one, the closing bracket is also written on a new line

Polyakiv-Andrey (changes requested):

app/main.py
AndriZhok: lol :/

Polyakiv-Andrey (approved):
Good job!

PR #1095, opened 2024-06-21
Develop

Wallian169:


app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+        if not coords:
+            self.coords = [0, 0]
```
vsmutok, line 12: better to use `or`  in this case

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        if not coords:
+            self.coords = [0, 0, 0]
```
vsmutok, line 34: better to use `or`  in this case

vsmutok (changes requested):

PR #1097, opened 2024-06-24
Add solution

MRozdoba:


app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+        if self.coords is None:
+            self.coords = [0, 0]
```
vsmutok, line 12: better to use `or` in this case

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        if coords is None:
+            self.coords = [0, 0, 0]
```
vsmutok, line 34: better to use `or` in this case

vsmutok (changes requested):

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
...
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = coords or [0, 0, 0]
```
Dimosphen1, line 31: There is no need to redefine `self.coords` after initialization, do it earlier while passing `coords` as an argument
```suggestion
        super().__init__(name, weight, coords or [0, 0, 0])
```

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Dimosphen1, line 7: It would be better to place each argument including `self` on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability

Dimosphen1 (changes requested):

pavlopro (approved):
Good Job!

PR #1105, opened 2024-07-02
Solution

AlexanderLEmani:


app/main.py
```diff
@@ -1,5 +1,59 @@
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
vasylhnatiuk, line 2:
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] = None) -> None:
```

vasylhnatiuk: add everywhere

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,5 +1,69 @@
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        if coords is None:
```
vasylhnatiuk, line 35: use `or` here as well

vasylhnatiuk (changes requested):

vsmutok (approved):
Looks Good!

PR #1114, opened 2024-07-10
Solution

Alex-Plugin:


app/main.py
```diff
@@ -2,4 +2,75 @@ class Cargo:
...
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+
+        self.name = name
+        self.weight = weight
+        if coords is None:
```
pavlopro, line 17: Simplify this to `self.coords = coords or [0, 0]`

app/main.py
```diff
@@ -2,4 +2,75 @@ class Cargo:
...
+
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
```
pavlopro, line 48: The same as above, here

pavlopro (changes requested):

app/main.py
Alex-Plugin: Thank you! Done!!!

Polyakiv-Andrey (approved):
Good job!

PR #1117, opened 2024-07-10
Solution delivery drone task

viktor-77:


app/main.py
```diff
@@ -1,5 +1,68 @@
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self, name: str,
+            weight: int,
+            coords: List[int] = None
+    ) -> None:
+        coords = [0, 0, 0] if coords is None else coords
+        super().__init__(name, weight, coords)
```
Dimosphen1, line 42: There is no need to define `coords` variabe, you could pass value in the `__init__` directly
```suggestion
        super().__init__(name, weight, coords or [0, 0, 0])
```

app/main.py
```diff
@@ -1,5 +1,68 @@
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self, name: str,
+            weight: int,
+            coords: List[int] = None
+    ) -> None:
+        self.coords = [0, 0] if coords is None else coords
```
Dimosphen1, line 15: Simplify this condition without `is None` check or use `or` expression

Dimosphen1 (changes requested):

pavlopro (approved):
Good Job!

PR #1118, opened 2024-07-11
implemented delivery drone inheritance

MasakDirt:


app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
+            self.current_load = cargo
+
+    def unhook_load(self) -> None:
+        if self.current_load is not None:
```
oliasymak24, line 63: Use just `if self.current.load` will be enough.

oliasymak24 (approved):
In general your code looks good :)

PR #1122, opened 2024-07-13
Implemented solution

dimak20:


app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load:
+            print(f"didn't hook {cargo}, cargo already in current load")
+            return
+        if cargo.weight > self.max_load_weight:
+            print(f"didn't hook {cargo}, cargo is too heavy")
```
oliasymak24, line 64: I think it's bad just to use `cargo` in print message.

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int,
+                 coords: list = None) -> None:
+        if not coords:
+            coords = [0, 0, 0]
+        if len(coords) < 3:
```
oliasymak24, line 53: I'm not sure that you need this `if`.

oliasymak24 (changes requested):
In general your code looks but pay attention to some comments.

app/main.py
dimak20: Sorry, didn't notice that. There will be just object reference

oliasymak24 (approved):
Good job!

PR #1128, opened 2024-07-22
Solution

DSahalatyi:


app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
MNDonut, line 32: coords can be None as well

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: int,
+            coords: list = None,
```
MNDonut, line 51: coords can be None as well

MNDonut (changes requested):

PR #1133, opened 2024-07-25
'Solution'

RToister:


app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list | None = None
```
```suggestion
                 coords: list[int] | None = None
```

app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
+        self.coords[2] += step
+
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self, name: str,
+            weight: int, coords: list | None = None,
```
vasylhnatiuk, line 54: same here

app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] | None = None
+                 ) -> None:
+        if coords is None:
```
vasylhnatiuk, line 38: use or here as well

vasylhnatiuk (changes requested):

app/main.py
RToister: Thanks, will fixed

RToister:
Hello. I try to understand why I can't make one more pull request after first (  
for Solution3

PR #1136, opened 2024-07-27
implement py-delivery-drone

ValeriiaKyr:


app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
```
oliasymak24, line 11: `coords` can be also None, check in all places

app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
+    ) -> None:
+        if coords is None:
+            coords = [0, 0]
```
oliasymak24, line 14: you can use for self.coords ternary operator, check in all places

app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo,
```
oliasymak24, line 59: Can current_load be None?

oliasymak24 (changes requested):

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list | None = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
oliasymak24, line 15: For better readability will be better first to put the case when coords is None

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo | None,
```
oliasymak24, line 56: Also here I think you can put a default value for Cargo

oliasymak24 (changes requested):

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: None | Cargo = 0,
```
oliasymak24, line 56: Default value will be better to set None

oliasymak24 (approved):

PR #1138, opened 2024-07-28
Solution

Narberal90:


app/main.py
```diff
@@ -1,5 +1,78 @@
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: List[int] = None
```
oliasymak24, line 14: `coords` can be also None

app/main.py
```diff
@@ -1,5 +1,78 @@
...
+    ) -> None:
+        if coords is None:
```
oliasymak24, line 16: you can use ternary operator here, check in all places

app/main.py
```diff
@@ -1,5 +1,78 @@
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: List[int] = None
+    ) -> None:
+        if coords is None or len(coords) == 2:
```
oliasymak24, line 45: Do you need check for length?

app/main.py
```diff
@@ -1,5 +1,78 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        max_load_weight: int,
+        current_load: Cargo,
```
oliasymak24, line 62: Cargo can be None

app/main.py
```diff
@@ -1,5 +1,78 @@
...
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.current_load = current_load
+        self.max_load_weight = max_load_weight
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None:
+            self.current_load = (
+                cargo
+                if self.max_load_weight >= cargo.weight
```
oliasymak24, line 73: you can move this check to your fist `if`

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1,5 +1,74 @@
...
-# write your code here
+
+class BaseRobot:
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: List[int] = None
+    ) -> None:
+        self.coords = [0, 0] if coords is None else coords
```
LiudmylaKulinchenko, line 16: Let's check `if not coords` instead to cover the case where coords will be an empty list, same throughout the code

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 🔥

PR #1139, opened 2024-07-28
Solution

Ivan-Shakhman:
Solution of task

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2,4 +2,62 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords:
+            self.coords = coords
+        else:
+            self.coords = [0, 0]
+
+    def go_forward(self, step: int = 1) -> None:
+        self.coords[1] += step
+
+    def go_back(self, step: int = 1) -> None:
+        self.coords[1] -= step
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def go_right(self, step: int = 1) -> None:
+        self.coords[0] += step
+
+    def get_info(self) -> str:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
oliasymak24, line 7: `coords` can be also None, check in all places

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
+        self.name = name
+        self.weight = weight
+        if coords:
```
oliasymak24, line 10: you can use ternary operator here, check in all places

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None,
+            max_load_weight: int = 0,
+            current_load: Cargo = None
```
oliasymak24, line 52: Cargo can be None

oliasymak24 (changes requested):

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo | None) -> None:
+        if (
+                self.current_load is None
+                and cargo and cargo.weight <= self.max_load_weight
+        ):
```
LiudmylaKulinchenko, line 59: `has hook_load method taking Cargo object and saves it to current_load if current_load is None and cargo.weight not greater than max_load_weight of the drone` -- you don't need None in annotations in this method and u can remove `if cargo` check

LiudmylaKulinchenko (changes requested):

oliasymak24 (approved):

PR #1141, opened 2024-07-28
Solution for py-delivery-drone

AndriiHamasa:


app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
oliasymak24, line 7: `coords` can be also None, check in all places

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+        if coords is None:
```
oliasymak24, line 8: you can use ternary operator here, check in all places

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 current_load: int,
```
oliasymak24, line 49: bad annotation for `current_load`

oliasymak24 (changes requested):

app/main.py
oliasymak24: list | None = None will clearly indicate that it can be either a list or None

app/main.py
```diff
@@ -1,5 +1,61 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Union[int, None],
```
oliasymak24, line 49: Still incorrect, `current_load` can be Cargo or None, look more closely at the task

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1,5 +1,71 @@
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: Union[list, None] = None
```
oliasymak24, line 14: No need now to use Union, as new Python version support just list | None

oliasymak24 (approved):

PR #1142, opened 2024-07-28
'Solution'

RToister:


app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] | None = None
+                 ) -> None:
+        coords = coords or [0, 0, 0]
+        if len(coords) == 2:
+            coords.append(0)
```
vsmutok, line 40: why you need this two lines here?

vsmutok (changes requested):

app/main.py
RToister: To add vertical parameters needed for our robot to fly.

PR #1143, opened 2024-07-28
Solution

Bogdan-Rylov:


app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
oliasymak24, line 7: `coords` can be also None

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+        if not coords:
```
oliasymak24, line 8: you can use ternary operator here, check in all places

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo = None,
```
oliasymak24, line 54: Cargo can be also None

oliasymak24 (changes requested):

app/main.py
```diff
@@ -5,11 +5,9 @@ def __init__(self, weight: int) -> None:
 
 class BaseRobot:
     def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
MNDonut, line 7: still not fixed

app/main.py
```diff
@@ -5,11 +5,9 @@ def __init__(self, weight: int) -> None:
...
-        if not coords:
-            coords = [0, 0]
         self.name = name
         self.weight = weight
-        self.coords = coords
+        self.coords = coords if coords is not None and coords else [0, 0]
```
MNDonut, line 10: `if coords` is enough

app/main.py
```diff
@@ -34,8 +32,11 @@ def __init__(
             weight: int,
             coords: list = None
```
MNDonut, line 33: coords can be None

app/main.py
```diff
@@ -34,8 +32,11 @@ def __init__(
...
     ) -> None:
-        if not coords or len(coords) != 3:
-            coords = [0, 0, 0]
+        coords = (
+            coords
+            if coords is not None and len(coords) == 3
```
MNDonut, line 37: `not coords` is enough instead of None check

app/main.py
MNDonut: still not fixed

MNDonut (changes requested):
Some issues still aren't fixed + new issues

app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
+            current_load: Cargo | None = None,
+            coords: list | None = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = (
+            current_load
+            if isinstance(current_load, Cargo)
+            else None
+        )
```
LiudmylaKulinchenko, line 64: You don't need it. If you have default value, it will give you None anyway

LiudmylaKulinchenko (changes requested):

app/main.py
Oleksl888: You should annotate that coords is an optional parameter - same as you did in below classes. Also you need to put parameters each on new line

Oleksl888 (changes requested):

PR #1144, opened 2024-07-29
Solution

pashawarganov:


app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
oliasymak24, line 7: `coords` can be also None

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
+        self.name = name
+        self.weight = weight
+        if coords is None:
```
oliasymak24, line 10: You can use ternary operator here, check in all places

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo,
```
oliasymak24, line 51: Cargo can be None

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list | None = None
+    ) -> None:
+        super().__init__(name, weight, [0, 0, 0]) if coords is None \
+            else super().__init__(name, weight, coords)
```
oliasymak24, line 44: no need to call the super function twice
`coords` can be defined first then pass to function

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo | None,
```
oliasymak24, line 59: Default value for Crago will be also good to define

oliasymak24 (changes requested):

oliasymak24 (approved):
Good job!

PR #1145, opened 2024-07-29
Soluton

mihavryliuk:


app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
oliasymak24, line 7: `coords` also can be None

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+        if coords is None:
```
oliasymak24, line 8: You can use ternary operator here, check in all places

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 coords: list = None,
+                 current_load: Cargo = None) -> None:
```
oliasymak24, line 49: Cargo also can be None

oliasymak24 (changes requested):

PR #1146, opened 2024-07-29
first_try

yashkunn:


app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot(Cargo):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0]
+        super().__init__(weight)
+        self.name = name
+        self.coords = coords
+        self.coords = coords if coords else [0, 0]
```
oliasymak24, line 13: Why you need two `if` for coords?

app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
+    #     directions = {
+    #         "forward": (0, step),
+    #         "back": (0, -step),
+    #         "right": (step, 0),
+    #         "left": (-step, 0)
+    #     }
+    #     if direction in directions:
+    #         self.coords[0] += directions[direction][0]
+    #         self.coords[1] += directions[direction][1]
+    # why isn't it working?
```
oliasymak24, line 37: Delete this commented out code, this is not good practice to put questions here, ask in chat or in Qna)

app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot(Cargo):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
oliasymak24, line 7: `coords` can be not only a list, also None, check in all places

app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None,
+            max_load_weight: int = 0,
+            current_load: Cargo = None,
```
oliasymak24, line 61: `current_load` can be also None

app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
+            self.current_load = cargo
+    # tried to use the ternary operator to assign,
+    # but it looked terrible
```
oliasymak24, line 71: This approach looks correct

app/main.py
oliasymak24: This move function must work correctly
What problem exactly do you have?

oliasymak24 (changes requested):

app/main.py
yashkunn: I get an error that BaseRobot must have methods go_forward, go_back...

oliasymak24: yes, because your tests expect exactly those function names that are in the task, but in general this function will work if there are no explicit function requirements

app/main.py
oliasymak24: You still didn't change it

app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
+            current_load: None | Cargo = None,
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
+            self.current_load = cargo
+    # tried to use the ternary operator to assign,
```
oliasymak24, line 67: delete this comments too

oliasymak24 (changes requested):

app/main.py
yashkunn: sorry for the inattention

LiudmylaKulinchenko (approved):
Good job 👏

PR #1148, opened 2024-07-29
Solution by Politov

Alexandr-Politov:


app/main.py
```diff
@@ -1,5 +1,73 @@
...
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = [0, 0] if coords is None else coords
```
LiudmylaKulinchenko, line 13: Let's check `if not coords` instead to cover the case where coords will be an empty list

LiudmylaKulinchenko: same everywhere

app/main.py
```diff
@@ -1,5 +1,73 @@
...
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
```
LiudmylaKulinchenko, line 34: Use ternary operator also here to be consistent

app/main.py
```diff
@@ -1,5 +1,73 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None,
+                 max_load_weight: int = 0,
+                 current_load: int = None) -> None:
```
LiudmylaKulinchenko, line 50: Is current_load actually int?

app/main.py
```diff
@@ -1,5 +1,73 @@
...
+drone = DeliveryDrone(
+    name="Jim",
+    weight=18,
+    coords=[11, -4, 16],
+    max_load_weight=20,
+    current_load=None,
+)
+drone.hook_load(cargo)
+print(drone.current_load)
+# drone.current_load is cargo
```
LiudmylaKulinchenko, line 73: remove unnecessary comment

app/main.py
```diff
@@ -1,5 +1,73 @@
...
+cargo = Cargo(14)
+drone = DeliveryDrone(
+    name="Jim",
+    weight=18,
+    coords=[11, -4, 16],
+    max_load_weight=20,
+    current_load=None,
+)
+drone.hook_load(cargo)
+print(drone.current_load)
```
LiudmylaKulinchenko, line 72: Remove testing from the pr

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 🔥

PR #1149, opened 2024-07-29
first commit

panichqa:


app/main.py
```diff
@@ -2,4 +2,71 @@ class Cargo:
...
+class BaseRobot:
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: list = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
```
oliasymak24, line 15: You can use here ternary operator, check in all places

app/main.py
```diff
@@ -2,4 +2,71 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: list = None
```
oliasymak24, line 11: `coords` can be also None, not only list

app/main.py
```diff
@@ -2,4 +2,71 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: list = None,
+        max_load_weight: int = 0,
+        current_load: Cargo = None
```
oliasymak24, line 61: `current_load` also can be None

app/main.py
```diff
@@ -2,4 +2,71 @@ class Cargo:
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: list = None
+    ) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
```
oliasymak24, line 44: What if coords is not None?

oliasymak24 (changes requested):

app/main.py
oliasymak24: list | None = None will clearly indicate that it can be either a list or None

app/main.py
oliasymak24: Cargo | None = None will clearly indicate that it can be either a Cargo or None

oliasymak24 (changes requested):
Some of your tests were failed :)

app/main.py
```diff
@@ -1,5 +1,74 @@
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: list = None
```
Oleksl888, line 41: Annotate that `coords` is an optional parameter here and in below class

Oleksl888 (changes requested):

app/main.py
panichqa: I hope I got it right and corrected in the right places

oliasymak24 (approved):

PR #1150, opened 2024-07-29
Solution

FalseCoreCentury:


app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords : list = None) -> None:
```
EdAlekseiev, line 7: You should specify inner type of list in type hint. Also, if argument can be None, you should show it in type hint. Fix everywhere, please
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] | None = None) -> None:
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #1151, opened 2024-07-29
Task Solution

achalaidiuk:


app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
```
oliasymak24, line 11: `coords` also can be None, not just a list, check in all places

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
oliasymak24, line 15: For better readibility to put first `if coords is None`  will better

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None,
+            max_load_weight: int = 0,
+            current_load: Cargo = None
```
oliasymak24, line 57: `current_load` here also can be None

oliasymak24 (changes requested):

oliasymak24 (approved):
Good job!

PR #1152, opened 2024-07-29
Solution

mnezbrytska:


app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0]
```
oliasymak24, line 9: You can use ternary operator, check in other places

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
oliasymak24, line 7: `coords` also get be None not only list
list | None = None will clearly indicate that it can be either a list or None
check in other places

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
+
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load:
+            Cargo = None,
```
oliasymak24, line 51: `Cargo` must be in the same line with `current_load`
`current_load` can be not only a list but also None

oliasymak24 (changes requested):

app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list | None = None
+    ) -> None:
+        coords = coords if coords is not None else [0, 0, 0]
```
Oleksl888, line 41: No need to store it in a varibale - you can pass this expression to `__init__` as is

Oleksl888 (approved):
Good Job

Oleksl888 (approved):
Well done!

PR #1153, opened 2024-07-29
task done

Sergi0bbb:


app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
```
Oleksl888, line 11: Signify with type annotation that ` coords` is an optional parameter

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
```
Oleksl888, line 38: add annotation to show that coords is optional parameter

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+    ) -> None:
+        if coords is None:
```
Oleksl888, line 40: Use ternary operator or use expression with `or` to check and set `coords` in super call below

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: int,
```
Oleksl888, line 57: use correct type here, also mention optional parameters in this method as well where necessary

Oleksl888 (changes requested):

LiudmylaKulinchenko (approved):
Good job 👏

PR #1155, opened 2024-07-30
'Solution'

ABaidin:


app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Oleksl888, line 7: Write each  param from a new line, also you need to annotate that coords parameter is optional

app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
...
+        self.name = name
+        self.weight = weight
+        if not coords:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
Oleksl888, line 13: Simplify these lines using ternary operator or make one line expression with `or`

app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
...
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
```
Oleksl888, line 36: Add annotation that `coords` is optional

app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
...
+    ) -> None:
+        super().__init__(name, weight, coords)
+        if not coords:
```
Oleksl888, line 39: It would be better to pass correct `coords` to `__init__` straight away instead of reassigning them. Use one-line expression in a super call

app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo = None,
```
Oleksl888, line 57: Add annotations for optional parameters where necessary

app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
...
+            coords: list = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        if not coords:
```
Oleksl888, line 61: This class inherits from FlyingRobot which accepts same parameters so why do you need to perform this check here?

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: Optional[list] = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = [0, 0, 0] if not coords else coords
```
Oleksl888, line 44: why don't you call `super().__init__` already with correct parameters?

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: Optional[list] = None
+    ) -> None:
+        coords = [0, 0, 0] if not coords else coords
```
Oleksl888, line 43: you can put this expression in super() call below so you don't need to store an extra variable

Oleksl888 (changes requested):

Oleksl888 (approved):
Bravo :clap:

PR #1156, opened 2024-07-30
add 3 class BaseRobot, FlyingRobot and DeliveryDrone

egor250330:


app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: list[int] = None
```
Oleksl888, line 11: Add type annotation that coords parameter is optional

app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
+    ) -> None:
+        if coords is None:
```
Oleksl888, line 13: Simplify these lines using ternary operator or make one line expression with `or`

app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: list[int] = None
```
Oleksl888, line 40: Annotate that coords is optional parameter - check other classes/methods and add where necessary

app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
+    ) -> None:
+        if coords is None:
```
Oleksl888, line 42: You can pass one-line expression setting correct `coords` to `__init__` straight away

app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        max_load_weight: int,
+        current_load: Cargo = None,
+        coords: list[int] = None
+    ) -> None:
+        if coords is None:
```
Oleksl888, line 62: This will be already implemented in parent class so why would you repeat yourself?
Also please check annotations here

app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None:
+            if cargo.weight <= self.max_load_weight:
```
Oleksl888, line 70: These can be combined in one `if` statement

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,5 +1,71 @@
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: Optional[list[int]] = None
```
MNDonut, line 14: after 3.10 you can use | operator to list optional parameters

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+    ) -> None:
+        coords = coords if coords is not None else [0, 0]
```
MNDonut, line 16: let's do it straightforward and use if coords condition

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        max_load_weight: int,
+        current_load: Cargo = None,
```
MNDonut, line 59: cargo can be None

MNDonut (changes requested):

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo | None = None,
+            coords: list[int] = None
```
oliasymak24, line 60: `coords` can be None here also too, not only list

app/main.py
```diff
@@ -1,5 +1,71 @@
...
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list | None = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
oliasymak24, line 19: Ihor left you comment about this, "let's do it straightforward and use if coords condition"

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1162, opened 2024-07-30
Task 11

Poznanskyi:


app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: str,
```
LiudmylaKulinchenko, line 47: Is current load string?

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+            coords: list = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: str) -> None:
```
LiudmylaKulinchenko, line 54: `has hook_load method taking Cargo object and saves it to current_load if current_load is None and cargo.weight not greater than max_load_weight of the drone` - it sayis that function takes Cargo object, not string

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+
+    def get_info(self) -> None:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
```
LiudmylaKulinchenko, line 32: You should use ternary operator here as well to be consistent throughout your code

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
LiudmylaKulinchenko, line 10: Let's just do `if not coords` instead, if will also cover a case where coords in an empty list

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
...
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> None:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        coords = coords if coords is not None else [0, 0, 0]
```
Oleksl888, line 30: rewrite this as per Liudmila's comment

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
```
Oleksl888, line 8: It is a good practice to break many params so they start each at the new line

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: None,
```
Oleksl888, line 46: You may also expecting Cargo object here

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Oleksl888, line 7: You should also mention in annotations that some parameters are optional. `coords, current_load` are such parameters

Oleksl888 (changes requested):

app/main.py
oliasymak24, line 7: Check it once again, coords can be also None, not just a list,  list | None = None

app/main.py
oliasymak24: Check this too

app/main.py
oliasymak24, line 46: current_load can be Cargo or None

oliasymak24 (changes requested):

LiudmylaKulinchenko (approved):
I don't see changes after the last review 👀

PR #1163, opened 2024-07-30
Update main.py

DaniaK123:


app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
vsmutok, line 10: better to use `or` in this case

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+        self.coords[0] -= step
+
+    def get_info(self) -> None:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
```
vsmutok, line 31: better to use `or` in this case

vsmutok (changes requested):

PR #1195, opened 2024-08-31
Develop

oksanaklw:


app/main.py
```diff
@@ -1,5 +1,56 @@
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
vsmutok, line 5: better to use `or` in this case

app/main.py
```diff
@@ -1,5 +1,56 @@
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
```
vsmutok, line 27: better to use `or` in this case

vsmutok (changes requested):

app/main.py
```diff
@@ -1,5 +1,55 @@
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
vasylhnatiuk, line 2:
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] = None) -> None:
```

vasylhnatiuk (changes requested):

PR #1197, opened 2024-09-01
Solution

PonomaryovNikita:


app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
...
+
+class BaseRobot(Cargo):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None
+                 ) -> None:
+        super().__init__(weight)
+        self.name = name
+        if coords is None:
```
pavlopro, line 14: Simplify to  

app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        super().__init__(name, weight, coords)
```
pavlopro, line 42: simplify to` super().__init__(name, weight, coords or [0, 0, 0])`

app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None,
+            max_load_weight: int = 0,
+            current_load: int = None
```
pavlopro, line 62: Correct annotation for current_load
You should cover both Cargo and None in annotations for current_load

pavlopro (changes requested):

app/main.py
pavlopro: `Cargo | None = None`

pavlopro (changes requested):

app/main.py
pavlopro: ` current_load: Cargo | None = None`

pavlopro (changes requested):

PR #1199, opened 2024-09-02
implemented 3 classes: BaseRobot, FlyingRobot, DeliveryDrone

Yurii-Durden:


app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
```
vasylhnatiuk, line 11:
```suggestion
            coords: list[int] = None
```

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = [0, 0] if not coords else coords
```
vasylhnatiuk, line 15:
```suggestion
        self.coords = coords or  [0, 0] 
```

vasylhnatiuk (changes requested):

app/main.py
Yurii-Durden: +

vasylhnatiuk (approved):

PR #1203, opened 2024-09-06
Solution_1

apmaz:


app/main.py
```diff
@@ -1,5 +1,81 @@
...
+            weight: int,
+            coords: None | list = None
+    ) -> None:
+
+        if coords is None:
+            coords = [0, 0]
+
+        self.name = name
+        self.weight = weight
+        self.coords = coords
```
Dimosphen1, line 22: Avoid using if condition, perform assignment in one line using either if condition 
```suggestion
        self.name = name
        self.weight = weight
        self.coords = coords if coords else [0, 0]
```
or `or` expression
```suggestion
        self.name = name
        self.weight = weight
        self.coords = coords or [0, 0]
```

app/main.py
```diff
@@ -1,5 +1,81 @@
...
+            name: str,
+            weight:
+            int,
+            coords: None | list = None
+    ) -> None:
+
+        if coords is None:
+            coords = [0, 0, 0]
+
+        super().__init__(name=name, weight=weight, coords=coords)
```
Dimosphen1, line 52: Perform the assignment directly in the `super().__init__(`
```suggestion
        super().__init__(name=name, weight=weight, coords=coords or [0, 0, 0])
```

Dimosphen1 (changes requested):

pavlopro (approved):
GJ!

PR #1208, opened 2024-09-15
task ready

Lev2098:
add class BaseRobot(Cargo), FlyingRobot(BaseRobot), DeliveryDrone(FlyingRobot)
add method go_ + forward, back, right, left, up, down
add hook load and unhook_load

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2,4 +2,75 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot(Cargo):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        super().__init__(weight)
+        if coords is None:
+            coords = [0, 0]
+        self.name = name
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -2,4 +2,75 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot(Cargo):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] = None) -> None:
```

app/main.py
```diff
@@ -2,4 +2,75 @@ class Cargo:
...
+        super().__init__(weight)
+        if coords is None:
```
vasylhnatiuk, line 9: do it in one line and use `or`

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1212, opened 2024-09-19
Develop

KKvladys:


dmytrik (approved):
Screenshot 2024-09-21 at 16 32 59
I think we need to send coords flyingRobots to parent class. But i'm not sure.

AnyoneClown (approved):
Good job!

PR #1217, opened 2024-09-25
Solution

Atikiho:


app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        self.coords = coords if coords is not None else [0, 0, 0]
+        super().__init__(name, weight, self.coords)
```
AnyoneClown, line 41: We can refactor it a bit.

```
```

If `coords` - None, it will use list with 0s. Check all cases

AnyoneClown (changes requested):

Nattalli (approved):

PR #1219, opened 2024-09-26
create classes Cargo, BaseRobot, FlyingRobot, DeliveryDrone and their methods

Mxbely:


Atikiho (approved):
LGTM

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Nattalli, line 7: specify `list` type

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Nattalli, line 29: and here

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
...
+        super().__init__(name, weight)
+        self.coords = [0, 0, 0] if coords is None else coords
```
Nattalli, line 31: can you send `coords` variable to parents `__init__`?

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo,
+            coords: list = None
```
Nattalli, line 47: and here

Nattalli (changes requested):

app/main.py
Mxbely: yes, it can be image like that: `super().__init__(name, weight, [0, 0, 0] if coords is None else coords)`

AnyoneClown (approved):
LGTM!

PR #1221, opened 2024-10-01
Solution

VladyslavBon:


app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords else [0, 0]
```
Atikiho, line 15: You can use `coords or [0, 0]` in all cases. It will use [0, 0] if coords equals None

Atikiho (changes requested):

AnyoneClown (approved):
LGTM!

PR #1222, opened 2024-10-01
Solution

vladyslav-tmf:


app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
```
dmytrik, line 30: you can use  coords=coords or [0, 0, 0] without if

dmytrik (approved):

app/main.py
vladyslav-tmf: I know it's not necessary, but I decided to add this check in case the user entered some coords. Although now I realized that I could have just typed coodrs = coords. Thanks)

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
AnyoneClown, line 10: We can use `or` here.

app/main.py
AnyoneClown: +, fix this 👍

app/main.py
```diff
@@ -2,4 +2,58 @@ class Cargo:
...
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None,
```
AnyoneClown, line 46: We can add more descriptive `coords` type hint. `list[int]`, for example

AnyoneClown (changes requested):

app/main.py
vladyslav-tmf: > +, fix this 👍

Done

app/main.py
vladyslav-tmf: > We can use `or` here.

+

AnyoneClown (approved):
Good job!

PR #1224, opened 2024-10-02
implement py-delivery-drone

yuliiapyrih:


app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
LiudmylaKulinchenko, line 10: You can just check `if coords` here. It will also cover the cases where coords will be a blank list

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo,
```
LiudmylaKulinchenko, line 46: current_load could also be none

LiudmylaKulinchenko (changes requested):

app/main.py
yuliiapyrih: Done

app/main.py
yuliiapyrih: Done

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
```
Nattalli, line 8: you can specify here:

```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] = [0, 0]) -> None:
        self.name = name
```

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
...
+        self.weight = weight
+        self.coords = coords if coords else [0, 0]
```
Nattalli, line 10: so you don't need here `else [0, 0]` if you specify it in annotation
```suggestion
        self.coords = coords
```

app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
...
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.coords = coords if coords else [0, 0, 0]
+        super().__init__(name, weight, self.coords)
```
Nattalli, line 31:
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] = [0, 0, 0]) -> None:
        super().__init__(name, weight, coords)
```

Nattalli (changes requested):

app/main.py
yuliiapyrih: if I fix the code as you indicated, the tests will fail

app/main.py
yuliiapyrih: if I fix the code as you indicated, the tests will fail

app/main.py
yuliiapyrih: if I fix the code as you indicated, the tests will fail

Nattalli (approved):

PR #1230, opened 2024-10-04
'Solution'

spa7id:


app/main.py
```diff
@@ -1,5 +1,76 @@
...
+                 name: str,
+                 weight: int,
+                 coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
Nattalli, line 18:
```suggestion
        self.coords = coords or [0, 0]
```

app/main.py
```diff
@@ -1,5 +1,76 @@
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
```
Nattalli, line 46: fix here too

app/main.py
```diff
@@ -1,5 +1,76 @@
...
+                 coords: list = None,
+                 ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
+            self.current_load = cargo
+        return self.current_load
```
Nattalli, line 73: shouldn't return anything

Nattalli: as you mentioned `-> None`

Nattalli (changes requested):

app/main.py
Nattalli: nooo, just remove `return ...`

Nattalli (approved):

PR #1239, opened 2024-10-07
solution

pohgen:


app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if not coords:
+            coords = [0, 0]
```
molodsh1y1, line 9: I think better used `self.coords = coords or [0, 0]`

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
...
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if not coords:
+            coords = [0, 0, 0]
```
molodsh1y1, line 33: The same

molodsh1y1 (approved):

app/main.py
```diff
@@ -2,4 +2,59 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
```suggestion
    def __init__(self, name: str, weight: int, coords: list[int] = None) -> None:
```

app/main.py
Nattalli: agree

Nattalli (changes requested):

Nattalli (approved):

PR #1253, opened 2024-10-08
Solution

FursykIvan:


app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int, coords:
+                 list[int] = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
Rabb1t-Artem, line 13: change 'if" method to "or" method

app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
...
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        elif len(coords) == 2:
+            coords.append(0)
+        super().__init__(name, weight, coords)
```
Rabb1t-Artem, line 40: you can use more simple logic.

Rabb1t-Artem (changes requested):

app/main.py
Rabb1t-Artem: its about "if coords is not None else"

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        coords = coords or [0, 0, 0]
+        if len(coords) == 2:
+            coords.append(0)
+        super().__init__(name, weight, coords)
```
AnyoneClown, line 39:
```suggestion
        super().__init__(name=name, weight=weight, coords=coords)
        self.coords = coords or [0, 0, 0]
```

We can refactor it like this. When you pass coords to your `father` class with `super()` it will do the same logic

AnyoneClown (approved):
Good, but fix small mistake

PR #1255, opened 2024-10-08
Solution

Rabb1t-Artem:


app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+        coords: list[int] = None,
+        max_load_weight: int = 0,
+        current_load: Cargo | None = None,
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
```
AnyoneClown, line 58: We can refactor using `not` here

```suggestion
        if not self.current_load and cargo.weight <= self.max_load_weight:
```

AnyoneClown (approved):
Good, one small mistake, take a look

PR #1316, opened 2024-11-25
Created all methods and classes

shafran-andriy:


app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] | None = None
+                 ) -> None:
+        coords = [0, 0] if coords is None else coords
```
vasylhnatiuk:
```suggestion
        coords = coords or [0, 0]
```

app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+
+    def go_forward(self, step: int = 1) -> list:
```
vasylhnatiuk:
```suggestion
    def go_forward(self, step: int = 1) -> None:
```

vasylhnatiuk: fix it everywhere

vasylhnatiuk (changes requested):

app/main.py
shafran-andriy: fixed

app/main.py
shafran-andriy: fixed

vsmutok (approved):
Looks Good!

PR #1330, opened 2024-11-30
Solution

julia4406:
About FlyingRobot
It doesn't work with 'isinstance', have used straight comparing with 'None'

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -4,13 +4,15 @@ class Cargo:
 
 
 class BaseRobot:
-    def __init__(self,
-                 name: str,
-                 weight: int,
-                 coords: list = None) -> None:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list | None = None
+    ) -> None:
         self.name = name
         self.weight = weight
-        self.coords = coords if coords is not None else [0, 0]
+        self.coords = coords or [0, 0]
 
     def go_forward(self, step: int = 1) -> None:
         self.coords[1] += step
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None) -> None:
```
L1nk3rrr: `coords` can be list or None, must be at annotation, will looks like `coords: list | None = None`

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
L1nk3rrr: no sense to use not here, simply swap an arguments.
Also, you can use simplified version like `self.coords = cords or [0, 0]`

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 coords: list = None,
+                 current_load: Cargo = None) -> None:
```
L1nk3rrr: same, can be None at annotation

app/main.py
```diff
@@ -2,4 +2,62 @@ class Cargo:
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None) -> None:
+        if coords is None or len(coords) < 3:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
```
L1nk3rrr: I think you can simplify it like on your `BaseRobot`, at task no need to check len of coords

L1nk3rrr (changes requested):
Nice, logically all is accepted by task requirements, minor styling, and annotation points.
Also, look at the checklist, and use this practice, for example -> If the function definition line is too long, place each parameter on a new line.

L1nk3rrr (approved):
Well done! Great work Julia!

PR #1331, opened 2024-12-01
Solution

MarianaAB19:


app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            coords: list = [0, 0, 0],
+            current_load: "Cargo" = None
+    ) -> None:
```
vsmutok: Using a list as the default value for the `coords` argument can lead to unexpected results, as all instances of the class will share the same list. It is better to use `None` as the default value and initialize the list inside the constructor.

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #1345, opened 2024-12-05
Delivery drones instances

valerii-kashpur:
- added BaseRobot class
- added FlyingRobot class
- added DeliveryDrone class

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -8,7 +8,7 @@ class BaseRobot:
             self,
             name: str,
             weight: int,
-            coords: list = None
+            coords: list | None = None
     ) -> None:
         self.name = name
         self.weight = weight
@@ -35,7 +35,7 @@ class FlyingRobot(BaseRobot):
             self,
             name: str,
             weight: int,
-            coords: list = None
+            coords: list | None = None
     ) -> None:
         if not coords:
             coords = [0, 0, 0]
@@ -54,11 +54,9 @@ class DeliveryDrone(FlyingRobot):
             name: str,
             weight: int,
             max_load_weight: int,
-            current_load: Cargo = None,
-            coords: list = None,
+            current_load: Cargo | None = None,
+            coords: list | None = None,
     ) -> None:
-        if not coords:
-            coords = [0, 0, 0]
         super().__init__(name, weight, coords)
         self.max_load_weight = max_load_weight
         self.current_load = current_load
@@ -66,9 +64,8 @@ class DeliveryDrone(FlyingRobot):
     def hook_load(self, cargo: Cargo) -> None:
         max_load_weight = self.max_load_weight
         current_load = self.current_load
-        can_hook = cargo.weight <= max_load_weight and not current_load
 
-        if can_hook:
+        if cargo.weight <= max_load_weight and not current_load:
             self.current_load = cargo
 
     def unhook_load(self) -> None:
--- a/app/main.py
+++ b/app/main.py
@@ -2,4 +2,74 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords else [0, 0]
+
+    def go_forward(self, step: int = 1) -> None:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        max_load_weight = self.max_load_weight
+        current_load = self.current_load
+        can_hook = cargo.weight <= max_load_weight and not current_load
+
+        if can_hook:
```
L1nk3rrr: looks like you don't need an extra variables and you can check it at `if` simply

app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
```
L1nk3rrr: it can be list or None

app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
```
L1nk3rrr: it can be list or None

app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo = None,
+            coords: list = None,
```
L1nk3rrr: it can be list or None

app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo = None,
```
L1nk3rrr: it can be Cargo or None

app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
+            coords: list = None,
+    ) -> None:
+        if not coords:
+            coords = [0, 0, 0]
```
L1nk3rrr: If `DeliveryDrone` is a child of `FlyingRobot`, do you need this part of code?

L1nk3rrr (changes requested):
Well done! Please improve your code due to commends

app/main.py
valerii-kashpur: just because don't like long if condition expressions

app/main.py
```diff
@@ -2,4 +2,71 @@ class Cargo:
...
+            current_load: Cargo | None = None,
+            coords: list | None = None,
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        max_load_weight = self.max_load_weight
+        current_load = self.current_load
```
L1nk3rrr: you don't need external variables with same names as your attributes

L1nk3rrr: if you don't like long if statements, okey, you can save your previous variable, but not use duplicated ones..
example:
```
        can_hook = cargo.weight <= self.max_load_weight and not self.current_load:
        if can_hook:
            self.current_load = cargo
```

or 

```
        if cargo.weight <= self.max_load_weight and not self.current_load:
            self.current_load = cargo
```

It looks like a normal if, not a long one

L1nk3rrr (changes requested):

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list | None = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords else [0, 0]
```
petrykivd, line 15: you can simplify it with "or"

petrykivd (approved):
Well done! 🚀

PR #1353, opened 2024-12-08
Solution

ruslanyemets:


app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list | None = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.coords = coords or [0, 0, 0]
```
4ikkamony: task description says to use conditioin to send correct coords, this works, too, of cource, but requires a bit more steps. First you pass unverifed coords to parent, it sets it to something and then you overwrite it anyway

app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+            self,
+            name: str,
+            weight: int,
+            coords: list | None = None,
+            max_load_weight: int = 0,
+            current_load: Cargo | None = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load or None
```
4ikkamony: IDK if I understood the task correctly, but there was a line "If Cargo object was passed to function, use method hook_load to check if it can be hooked."
I guess this method should be used.

4ikkamony (approved):
Nicely done!
Check those two lil' comments, everything else is perfect

PR #1356, opened 2024-12-09
solution

DanilTrukhanov:


maxakim0v (changes requested):
Use "flake8 app" with your terminal to see your mistake.

PR #1360, opened 2024-12-09
Solution

VovaGld:


app/main.py
```diff
@@ -1,5 +1,80 @@
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list | None = None
```
Dogober, line 14: I think to use two types list | None it looks weird. I think you need to indicate what exactly will be in the list like list[int] = None. Like that. But the best way will be to clarify it on Q&A

Dogober (approved):
Great work

PR #1361, opened 2024-12-09
'Solution'

Dogober:


app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            coords: list[int] = [0, 0, 0],
```
DanilTrukhanov: Better to made coords = None as a default argument, because if not it will end up that coords will be shared between all instances of a class.
https://www.pullrequest.com/blog/python-pitfalls-the-perils-of-using-lists-and-dicts-as-default-arguments/

DanilTrukhanov (approved):
Good!

app/main.py
Dogober: Thx for this correction!! And thx for usfull link

PR #1365, opened 2024-12-10
Solution

Sichkaridze:


MindDevastation (approved):

app/main.py
```diff
@@ -2,4 +2,75 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
petrykivd: coords if coords else [0,0] will be enough

app/main.py
```diff
@@ -2,4 +2,75 @@ class Cargo:
...
+
+    def go_forward(self, step: int = 1) -> None:
+        x, y, *_ = self.coords
+        y += step
+        self.coords[1] = y
```
petrykivd: You can simplify it like this:
self.coords[1] += step

petrykivd: In all places

app/main.py
```diff
@@ -2,4 +2,75 @@ class Cargo:
...
+                         coords if coords is not None else [0, 0, 0])
+
+        self.max_load_weight = max_load_weight
+        self.current_load = None
+        if current_load:
+            self.hook_load(current_load)
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
+            self.current_load = cargo
```
petrykivd: What happens if the object that gets here is not a cargo?

app/main.py
petrykivd: In all places

petrykivd (changes requested):

PR #1370, opened 2024-12-10
Solution

shretsker:


app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        if coords is None:
+            coords = [0, 0]
```
10kkyvl, line 14: You can do something like `coords = coords or [0, 0]`

10kkyvl (approved):

PR #1377, opened 2024-12-11
solution_v.1.0

yelizaveta-zh:


app/main.py
```diff
@@ -1,5 +1,67 @@
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 cargo: Cargo,
+                 coords: List[int] = None
+                 ) -> None:
```
Professor-Douel: Format your code like this:
```
def __init__(
        self,
        name: str,
        cargo: Cargo,
        coords: List[int] = None
) -> None:
```

or like this:
```
def __init__(self,
             name: str,
             cargo: Cargo,
             coords: List[int] = None) -> None:
```

app/main.py
```diff
@@ -1,5 +1,67 @@
...
+
+    def go_right(self, step: int = 1) -> None:
+        self.coords[0] += step
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return (f"Robot: {self.name}, "
+                f"Weight: {self.cargo.weight}")
```
Professor-Douel: You don`t need to split the string, string is not long.

app/main.py
```diff
@@ -1,5 +1,67 @@
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 cargo: Cargo,
+                 coords: List[int] = None
+                 ) -> None:
```
Professor-Douel: Need to format the code too.

app/main.py
```diff
@@ -1,5 +1,67 @@
...
+        self.current_load = None
+        if current_load:
+            self.hook_load(current_load)
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
+            self.current_load = cargo
+
+    def unhook_load(self) -> None:
+        self.current_load = None
```
Professor-Douel, line 71: No empty line after the code

app/main.py
```diff
@@ -1,5 +1,67 @@
...
+    def go_up(self, step: int = 1) -> None:
+        self.coords[2] += step
+
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class Drone(FlyingRobot):
+    def __init__(self, name: str, cargo: Cargo, max_load_weight: int,
+                 coords: List[int] = None, current_load: Cargo = None) -> None:
```
Professor-Douel: You need to format the code here too.

app/main.py
```diff
@@ -1,5 +1,67 @@
...
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 cargo: Cargo,
+                 coords: List[int] = None
```
Professor-Douel, line 13: You may change to:
coords: int = None
and than you don`t need the import

app/main.py
```diff
@@ -1,5 +1,67 @@
...
 class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 cargo: Cargo,
```
Professor-Douel: You change the name of the argument, but it does not match the condition of the task.
According to the conditions of the task, the BaseRobot class accepts the same arguments as the Cargo class.

Professor-Douel (changes requested):

app/main.py
Professor-Douel: you can rewrite:
`coords: list = None`
and you dont need to make import

Professor-Douel (changes requested):

app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
...
+class BaseRobot(Cargo):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        super().__init__(weight)
+        self.name = name
+        self.coords = coords if coords else [0, 0]
```
L1nk3rrr: you can simplify it to using `or`

app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
...
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            coords: list[int] = None,
+            current_load: Cargo = None
+    ) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
```
L1nk3rrr: you have this at your parent, no needed here

app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
...
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        self.coords = coords
+        super().__init__(name, weight, coords)
```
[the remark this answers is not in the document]
L1nk3rrr: he is right, because for now at your baserobot you don't have a lot of functionality, but in big project better to understand that re-assign better to do after __init__

L1nk3rrr (changes requested):
Minor comment for improvement!

app/main.py
yelizaveta-zh: how I can do it?

L1nk3rrr (approved):
Well done!

PR #1378, opened 2024-12-11
Task

Arty-tech-dotcom:


app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
L1nk3rrr: it can be written in one line. Also you can simplify it to usage of `or`

app/main.py
```diff
@@ -2,4 +2,65 @@ class Cargo:
...
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None,
+            max_load_weight: int = 0,
+            current_load: Cargo = None
+    ) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
```
L1nk3rrr: redundant `if` and default coords, it's already done at you Parent class `FlyingRobot`

L1nk3rrr (changes requested):

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None,
+            max_load_weight: int = 0,
+            current_load: Cargo = None
+    ) -> None:
+        coords = coords or [0, 0, 0]
```
L1nk3rrr, line 50: Such a check is already done at `FlyingRobot,` so you don't need it. Your parent class is doing this.

L1nk3rrr (changes requested):

PR #1397, opened 2024-12-20
all tests passed. flake8 passed

Mykola-Osolinskyi:


app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list | None = None
```
Nattalli: specify type of list

app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list | None = None,
```
Nattalli:
```suggestion
            coords: list[int] = [0, 0, 0],
```

app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
+            max_load_weight: int | None = None,
+            current_load: Cargo | None = None
+    ) -> None:
+        if not coords:
+            coords = [0, 0, 0]
```
Nattalli: so you can remove that now

app/main.py
```diff
@@ -2,4 +2,70 @@ class Cargo:
...
+        super().__init__(name, weight, coords)
```
Nattalli: or in that way

```suggestion
        super().__init__(name, weight, coords or [0, 0, 0])
```

Nattalli (changes requested):

app/main.py
Mykola-Osolinskyi: @Nattalli 
I did it according to this recommendation:
"It is not a good practice to use mutable object as a default parameter of function, set it to None by default. But in constructor itself you can use condition, and if coords equal to None, assign [0, 0] to the coords."

app/main.py
Mykola-Osolinskyi: @Nattalli 
this comment linked to comment above

Nattalli (approved):

PR #1411, opened 2025-01-05
done

Leo9siy:


app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+
+        if coords is None:
+            self.coords = [0, 0]
```
Dimosphen1, line 13: You could rewrite this line as follows
```suggestion
        self.coords = coords or [0, 0]
```

PR #1412, opened 2025-01-05
Solution

Baranotik15:


app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo = None,
```
oliasymak24, line 58: You use two different data types for these arguments, so use the correct type annotation current_load: Cargo | None = None. Fix in all other similar cases

app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
+        super().__init__(name=name, weight=weight, coords=coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = None
+        if current_load is not None:
+            self.hook_load(current_load)
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
+            self.current_load = cargo
+        return
```
oliasymak24: You don't need this return

app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+
+        self.coords = coords if coords is not None else [0, 0]
```
oliasymak24: Let's check ` if not coords` instead to cover the case where coords will be an empty list

oliasymak24 (changes requested):

app/main.py
```diff
@@ -8,10 +8,10 @@ def __init__(
             self,
             name: str,
             weight: int,
-            coords: list[int] = None
+            coords: list[int] | None = None
     ) -> None:
 
-        self.coords = coords if coords is not None else [0, 0]
+        self.coords = coords if coords else [0, 0]
```
Y-Havryliv, line 14: you can even use this:
```python
```

app/main.py
Y-Havryliv: you forgot to fix this one.

Y-Havryliv (approved):

PR #1449, opened 2025-01-25
'Solution'

nikita-plotnytskyi:


app/main.py
```diff
@@ -2,4 +2,75 @@ class Cargo:
...
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = [0, 0] if coords is None else coords
```
vsmutok: better to use `or` in this case

app/main.py
```diff
@@ -2,4 +2,75 @@ class Cargo:
...
+            name: str,
+            weight: int,
+            coords: list = None,
+            max_load_weight: int = 0,
+            current_load: Cargo = None
+    ) -> None:
+        super().__init__(
+            name=name,
+            weight=weight,
+            coords=[0, 0, 0] if coords is None else coords
```
vsmutok: better to use `or` in this case

vsmutok (changes requested):

vsmutok (approved):
Looks Good!

PR #1453, opened 2025-01-26
Solution for task delivery-drone

kram3ko:
"BaseRobot, FlyingRobot, and DeliveryDrone classes with movement and load handling functionality"

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2,4 +2,74 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None
+                 ) -> None:
+        if coords is None:
+            coords = [0, 0]
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__} {self.name}, Weight: {self.weight}"
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None
+                 ) -> None:
```
Y-Havryliv: image
try to format params like this. Or move this parenthesis on the line above

Y-Havryliv: fix everywhere

app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
+        if coords is None:
+            coords = [0, 0]
+        self.name = name
+        self.weight = weight
+        self.coords = coords
```
Y-Havryliv:
```suggestion
        self.coords = coords or [0, 0]
```
this one will be better. Also, by doing this you can delete the `if-statement` above. Fix everywhere.

app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.max_load_weight >= cargo.weight and self.current_load is None:
+            self.current_load = cargo
+
+    def unhook_load(self) -> None:
+        self.current_load = None
+
+
+if __name__ == "__main__":
```
Y-Havryliv: delete this code, you don't need it

Y-Havryliv (changes requested):

oliasymak24 (approved):

PR #1479, opened 2025-02-15
Finish

Sotnikov-100:


app/main.py
```diff
@@ -1,5 +1,74 @@
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        if coords is None:
+            coords = [0, 0]
```
oliasymak24: you can write this in one line.
self.coords = ...//put condition here too

app/main.py
```diff
@@ -1,5 +1,74 @@
...
-# write your code here
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None,
+            max_load_weight: int = 0,
+            current_load: Cargo = None
```
oliasymak24: You use two different data types for these arguments, so use the correct type annotation current_load: Cargo | None = None. Fix in all other similar cases

oliasymak24 (changes requested):

app/main.py
Sotnikov-100: Yes, it’s really useful

app/main.py
```diff
@@ -1,5 +1,70 @@
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] | None = None
+    ) -> None:
+        self.coords = coords if coords is not None else [0, 0]
```
Y-Havryliv, line 8:
```suggestion
        self.coords = coords or [0, 0]
```
it seems more simple

app/main.py
```diff
@@ -1,5 +1,70 @@
...
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] | None = None
+    ) -> None:
+        self.coords = coords if coords is not None else [0, 0, 0]
```
Y-Havryliv, line 35:
```suggestion
        self.coords = coords or [0, 0, 0]
```

Y-Havryliv (approved):
gj!

Y-Havryliv (approved):

PR #1496, opened 2025-02-23
'Solution'

AndyStarGitHub:


app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
```
iSevenpwnz, line 4: init Method Defined Outside a Class

app/main.py
```diff
@@ -2,4 +2,73 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list | None = None,
+                 max_load_weight : int = 0,
+                 current_load: Cargo = None) -> None:
```
iSevenpwnz, line 56: Undefined Type Cargo:
In the DeliveryDrone class, the constructor uses the type annotation Cargo for the current_load parameter.

However, there is no definition for Cargo in the code, which will lead to a NameError at runtime.

iSevenpwnz (changes requested):

app/main.py
AndyStarGitHub: Fixed

app/main.py
AndyStarGitHub: Fixed

app/main.py
```diff
@@ -1,5 +1,77 @@
...
```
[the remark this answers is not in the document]
AndyStarGitHub: Fixed

app/main.py
```diff
@@ -1,5 +1,77 @@
...
+        coords = coords or [0, 0, 0]
+        super().__init__(name,
+                         weight,
+                         [coords[0],
+                          coords[1],
+                          coords[2]])
```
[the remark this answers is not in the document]
AndyStarGitHub: Base class should have only two dimensions

PR #1566, opened 2025-03-26
Solution

olek007:


olek007:
@mateacademy-ai-mentor 
In the `DeliveryDrone` class, the `__init__` method first chheck whether `current_load` is a `Cargo` object or `None` and then calls `self.hook_load` depends on whether `current_load` is a `Cargo` object or `None`.

Line: 59:
` if isinstance(current_load, Cargo):
      self.hook_load(current_load)`

PR #1622, opened 2025-05-06
Solution

Nib20:


app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
```
MNDonut, line 15: `if not coords` would be enough

MNDonut (approved):

PR #1650, opened 2025-05-29
I figured out the imitations but the task is difficult

kravetst:


kravetb (changes requested):
Its look good but you can't explain this code!

PR #1661, opened 2025-06-05
Solution

vasyl-main-dev:


app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords is not None else [0, 0]
```
vsmutok:
```suggestion
        self.coords = coords or [0, 0]
```

app/main.py
```diff
@@ -2,4 +2,61 @@ class Cargo:
...
+        self.coords[0] -= step
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self, name: str, weight: int,
+            coords: list = None
+    ) -> None:
+        super().__init__(
+            name, weight, coords=[0, 0, 0]
```
vsmutok:
```suggestion
            name, weight, coords=None
```
It's not recommended to use mutable data types as default values.

vsmutok (changes requested):

app/main.py
vasyl-main-dev: Got it

app/main.py
vasyl-main-dev: Thanks, didn`t know that syntax

vasylhnatiuk (approved):

PR #1686, opened 2025-06-24
Solution

zqpiq:


app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Arch0998: This line contains 83 characters including spaces, but PEP 8 recommends a maximum of 79 characters.
Kindly edit this to comply with PEP 8 guidelines.

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
```
Arch0998: The same. This line contains 83 characters including spaces, but PEP 8 recommends a maximum of 79 characters.
Kindly edit this to comply with PEP 8 guidelines.

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+        if coords is None:
+            coords = [0, 0, 0]
+        elif len(coords) == 2:
```
Arch0998, line 45: What is this line used for?

app/main.py
Arch0998: You should use the following type annotation here:
`coords: Optional[List[int]] = None`
This clearly indicates that the variable coords can either be a list of integers or None. Using Optional helps to explicitly show that None is an acceptable default value, which improves code readability and makes it easier for static type checkers (like mypy) to verify your code for type safety. This also helps IDEs provide better autocompletion and reduces the chance of bugs related to unexpected types.

app/main.py
Arch0998: You should use the following type annotation here:
`coords: Optional[List[int]] = None`
This clearly indicates that the variable coords can either be a list of integers or None. Using Optional helps to explicitly show that None is an acceptable default value, which improves code readability and makes it easier for static type checkers (like mypy) to verify your code for type safety. This also helps IDEs provide better autocompletion and reduces the chance of bugs related to unexpected types.

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+            coords = coords + [0]
```
Arch0998, line 46: What is this line used for?

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int | float,
+            current_load: Cargo = None,
+            coords: list = None
+    ) -> None:
```
Arch0998, line 64: The closing parenthesis ) should be aligned with the start of the line where the def keyword begins.

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
```
Arch0998: `self.current_load = None`

app/main.py
```diff
@@ -2,4 +2,60 @@ class Cargo:
...
+
```
Arch0998, line 70: And after:
```
if current_load is not None:
    self.hook_load(current_load)
```
    
_This is specified in the condition_

Arch0998 (changes requested):
Please address the issues according to my comments.

app/main.py
zqpiq: Hi, this line adds a Z coordinate if necessary.

PR #1687, opened 2025-06-24
Solution

Arch0998:


app/main.py
```diff
@@ -1,5 +1,76 @@
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: Optional[List[int]] = None
+                 ) -> None:
```
AnyoneClown: Start every argument from a new line

app/main.py
```diff
@@ -1,5 +1,76 @@
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: Optional[List[int]] = None
+                 ) -> None:
```
AnyoneClown: Same here

app/main.py
```diff
@@ -1,5 +1,76 @@
...
+
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: Optional[List[int]] = None,
+                 max_load_weight: int = 0,
+                 current_load: Optional[Cargo] = None,
+                 ) -> None:
```
AnyoneClown: Fix here too

AnyoneClown (changes requested):

app/main.py
Arch0998: After def __init__(...) -> None: - a blank line is NOT needed, since the if statement is the first line of code in the method.
PEP 8 doesn't recommend adding blank lines at the beginning of functions/methods, only between logically different blocks within them.

app/main.py
Arch0998: After def __init__(...) -> None: - a blank line is NOT needed, since the if statement is the first line of code in the method.
PEP 8 doesn't recommend adding blank lines at the beginning of functions/methods, only between logically different blocks within them.

app/main.py
Arch0998: According to PEP 8, when there's a super() call in the __init__ method, an additional blank line after it is not needed.

Arch0998 (commented):
@AnyoneClown I disagree with you.
Both cases follow PEP 8's principle that blank lines should separate logically distinct sections of code, but within a single method like __init__, the entire body is typically considered one cohesive initialization block.

app/main.py
AnyoneClown: You can use 2 code styles. If you start `self` param at the same line with `(` - you should end your last param with `)` at the same line.

Example:
```suggestion
    def __init__(self,
                 name: str,
                 weight: int,
                 coords: Optional[List[int]] = None) -> None:
```

Example:
```suggestion
    def __init__(
        self,
        name: str,
        weight: int,
        coords: Optional[List[int]] = None
    ) -> None:
```

AnyoneClown (approved):
LGTM

PR #1694, opened 2025-06-27
Implemented a solution: created 3 classes that inherit from each other.

Otst0y:


app/main.py
```diff
@@ -2,4 +2,80 @@ class Cargo:
...
+
+        super().__init__(weight)
+        self.name = name
+        self.coords = coords or [0, 0]
+
+    def go_forward(self, step: int = 1) -> None:
+        self.coords[1] += step
+
+    def go_back(self, step: int = 1) -> None:
+        self.go_forward(-step)
```
dimashevchukk: this will cause an unnecessary method call, change it to using self.coords

app/main.py
```diff
@@ -2,4 +2,80 @@ class Cargo:
...
+
+    def go_right(self, step: int = 1) -> None:
+        self.coords[0] += step
+
+    def go_left(self, step: int = 1) -> None:
+        self.go_right(-step)
```
dimashevchukk: unnecessary method call

app/main.py
```diff
@@ -2,4 +2,80 @@ class Cargo:
...
 
-# write your code here
+
+class BaseRobot(Cargo):
+
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list | None = None
```
dimashevchukk: list[int] would be better

app/main.py
```diff
@@ -2,4 +2,80 @@ class Cargo:
...
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list | None = None
```
dimashevchukk: list[int]

app/main.py
```diff
@@ -2,4 +2,80 @@ class Cargo:
...
+    ) -> None:
+
+        super().__init__(name, weight)
+        self.coords = coords or [0, 0, 0]
```
dimashevchukk, line 44: can cause an error in future. prepare the coordinates, pass only two of them to the base class, and then assign the full value to the derived class, so coords [1,2,3] in flying robot will be [1,2] in regular, not [0,0]

app/main.py
```diff
@@ -2,4 +2,80 @@ class Cargo:
...
+
+    def go_up(self, step: int = 1) -> None:
+        self.coords[2] += step
+
+    def go_down(self, step: int = 1) -> None:
+        self.go_up(-step)
```
dimashevchukk: unnecessary method call

app/main.py
```diff
@@ -2,4 +2,80 @@ class Cargo:
...
+
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo = None,
+            coords: list | None = None,
```
dimashevchukk: list[int]

app/main.py
```diff
@@ -2,4 +2,80 @@ class Cargo:
...
+
+
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            current_load: Cargo = None,
```
dimashevchukk: annotation should be Cargo | None

app/main.py
```diff
@@ -2,4 +2,80 @@ class Cargo:
...
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+
+        if (
+            current_load is not None
+            and current_load.weight <= self.max_load_weight
+        ):
+            self.current_load = current_load
+        else:
+            self.current_load = None
```
dimashevchukk: no need for this, keep self.current_load = current_load

dimashevchukk (changes requested):

app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot(Cargo):
```
dimashevchukk, line 6: inheritance not needed

app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
...
+class FlyingRobot(BaseRobot):
+
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list[int] | None = None
+    ) -> None:
+        super().__init__(name, weight)
+        self.coords = coords or [0, 0, 0]
```
dimashevchukk, line 44: cords not passing to BaseRobot __init__ + ai comment

dimashevchukk (approved):

PR #1696, opened 2025-06-27
Solution

Sashoxxx:


app/main.py
```diff
@@ -1,5 +1,72 @@
...
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list | None = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords else [0, 0]
```
yagginov: With coords here it could be made simpler if instead of None as the default value you specify (0, 0) (i.e. tuple). And then in the body of `__init__` you simply write:
self.coords = list(coords)

This way you can avoid unnecessary checks

app/main.py
```diff
@@ -1,5 +1,72 @@
...
+        coords: list | None = None,
+    ) -> None:
+        super().__init__(name=name, weight=weight, coords=coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if self.current_load is None and cargo.weight <= self.max_load_weight:
+            self.current_load = cargo
+        return
```
yagginov: here `return` is unnecessary

yagginov (approved):
Everything is ok, except for non-critical moments

PR #1697, opened 2025-06-27
Solution for Delivery drone

yagginov:


Plovuezz (approved):
😍😍😍😍😍😍

yagginov:
@mateacademy-ai-mentor 
So in the constructor we should provide the ability to set current_load if it is provided, but only if the condition in hook_load is met. That is, if I do not write
`self.current_load = None`
in the constructor, then the `hook_load()` function will throw an error when trying to check `self.current_load is None`.
Therefore, I emphasize that my code does not violate any conditions and this was done specifically to avoid unnecessary checks and code complications.

PR #1700, opened 2025-06-27
Solution

mishahoshii:


app/main.py
```diff
@@ -2,4 +2,76 @@ class Cargo:
...
+            coords: list[int] | None = None
+    ) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+            parent_coords = [0, 0]
+        else:
+            parent_coords = coords[:2]
+
+        super().__init__(name, weight, parent_coords)
+        self.coords = coords
```
Vskesha: There is no need to set parent_coords because after that you replace the same variable `self.coords` to list of three integers.
My advice is:
```
        coords = coords or [0, 0, 0]
```
But decision is on you.

Vskesha (approved):
The code is working and meets all the requirements.

app/main.py
mishahoshii: Ok, thank you

PR #1711, opened 2025-06-29
Solution

SensibleN00B:


VadymKyrylko (changes requested):
Cargo is not defined as a class. There's only a constructor function.
Lack of name, coords in Cargo
hook_load() in DeliveryDrone.__init__  -> current_load is assigned directly without checking weight
Need to add "is None" guard for cargo in hook_load(), because If None is passed, cargo.weight will raise an error.

PR #1724, opened 2025-07-01
Solution

bohdankolisnyk04:


app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
+    def __init__(self, name: str, weight: int, coords : list = None) -> None:
+
+        if coords is None:
+            coords = [0, 0]
+
+        self.name = name
+        self.weight = weight
+        self.coords = coords
+
+    def go_forward(self, step : int = 1) -> None:
```
AnyoneClown, line 16:
```suggestion
    def go_forward(self, step: int = 1) -> None:
```

Refactor in all cases

AnyoneClown (approved):
Good, only code style fixes

PR #1727, opened 2025-07-02
Solution

Bebaaaaa:


app/main.py
```diff
@@ -1,5 +1,73 @@
...
+        self,
+        name: str,
+        weight: int,
+        coords: Optional[List[int]] = None,
+        max_load_weight: int = 0,
+        current_load: Optional[Cargo] = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load: Optional[Cargo] = None
```
L1nk3rrr, line 61: here looks like you need to set a value from init
```
self.current_load: Optional[Cargo] = current_load
```

app/main.py
```diff
@@ -1,5 +1,73 @@
+from typing import Optional, List
+
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int,
+                 coords: Optional[List[int]] = None) -> None:
```
L1nk3rrr, line 6: make the same styling as you did at DeliveryDrone, or if it can be one liner?

app/main.py
```diff
@@ -1,5 +1,73 @@
...
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int,
+                 coords: Optional[List[int]] = None) -> None:
```
L1nk3rrr, line 31: same here about styling

L1nk3rrr (changes requested):
Well done, but minor changes required to improve your code

PR #1845, opened 2025-10-21
solution of py-delivery-drone 1

KuroMidori123:


app/main.py
```diff
@@ -1,5 +1,87 @@
...
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list = None) -> None:
```
viktoria-rybenchuk: this argument can be None as well

app/main.py
```diff
@@ -1,5 +1,87 @@
...
+                 weight: int,
+                 coords: list | None = None) -> None:
+        if coords is None:
+            normalized_coords = [0, 0, 0]
+
+        elif len(coords) == 2:
+            normalized_coords = [coords[0], coords[1], 0]
+
+        elif len(coords) == 3:
+            normalized_coords = coords
```
viktoria-rybenchuk, line 45: why do you need this logic?

viktoria-rybenchuk (changes requested):

app/main.py
KuroMidori123: > why do you need this logic?

Чтобы всегда было 3 координаты

vsmutok (approved):
Great!

PR #1855, opened 2025-10-28
Added solution

vadym-zahrebelnyi:
The tests have once again expressed their profound dislike for documentation(

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,5 +1,77 @@
 class Cargo:
+
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+
+    def __init__(
+        self,
+        name: str,
+        weight: int,
+        coords: list[int] = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        self.coords = coords if coords else [0, 0]
+
+    def go_forward(self, step: int = 1) -> None:
+        self.coords[1] += step
+
+    def go_back(self, step: int = 1) -> None:
+        self.coords[1] -= step
+
+    def go_right(self, step: int = 1) -> None:
+        self.coords[0] += step
+
+    def go_left(self, step: int = 1) -> None:
+        self.coords[0] -= step
+
+    def get_info(self) -> str:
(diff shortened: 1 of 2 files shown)


app/main.py
```diff
@@ -1,5 +1,77 @@
...
+        self,
+        name: str,
+        weight: int,
+        coords: list[int] | None = None,
+        max_load_weight: int = 0,
+        current_load: Cargo | None = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = None
```
Vorongor, line 68: You describe that current load can be: current_load: Cargo | None = None. So we can put Cargo when create instance of DeliveryDrone. But prevent that possibility here.

app/main.py
```diff
@@ -1,5 +1,77 @@
...
+        if isinstance(current_load, Cargo):
+            self.hook_load(current_load)
```
Vorongor, line 70: Incorrect placement, you should verify current load before you set it to None.

app/main.py
```diff
@@ -1,5 +1,77 @@
...
+
+    def hook_load(self, cargo: Cargo) -> Cargo | None:
```
Vorongor: The type hint suggests it should return the Cargo object on success or None on failure, but it returns None always.

Vorongor (changes requested):
This is a well-structured set of classes that demonstrates solid Object-Oriented Programming (OOP) principles through inheritance! The design effectively models different types of robots by extending functionality from a base class. But have issue lies in the initialization logic of the DeliveryDrone class

app/main.py
vadym-zahrebelnyi: Yeap, it seems strange, but in Readme.md hook_load have to change state if both conditions are truth and do nothing otherwise, that's why i must define start state to None anyway,  or do it inside the function  that changes function's behavior (description say to do nothing if false)

app/main.py
vadym-zahrebelnyi: current_load as an arg can take Cargo, but current_load as an instance atribute don't have to. First we can create empty slot, then it docks smth if cond

PR #1860, opened 2025-10-30
Completed task

ostapshpeha:


app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
+                 name: str,
+                 weight: int,
+                 max_load_weight: int,
+                 coords: list | None = None,
+                 current_load: Cargo | None = None) -> None:
+        super().__init__(name, weight, coords=coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+        if current_load is not None:
+            self.hook_load(current_load)
```
vakt159, line 60: Can be replaced with:
if current_load is not None -> if current_load:

app/main.py
```diff
@@ -2,4 +2,69 @@ class Cargo:
...
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list | None = None) -> None:
+        if coords is None:
+            coords = [0, 0]
+        if len(coords) == 2:
+            coords.append(0)
+        super().__init__(name, weight, coords)
```
vakt159, line 40: Why don`t just write:
 if coords is None
     coords = [0, 0, 0]
  ?

vakt159 (changes requested):
Make minor changes and it will be good

app/main.py
ostapshpeha: Thanks

app/main.py
ostapshpeha: u have a ratio.
but condition 
```
if len(coords) == 2:
            coords.append(0)
```
is still important, if user inputs 2D coordinates in FlyingRobot class

vakt159 (approved):

PR #1861, opened 2025-10-30
Solution

vakt159:


app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
...
+    def go_down(self, step: int = 1) -> None:
+        self.coords[2] -= step
+
+
+class DeliveryDrone(FlyingRobot):
+
+    def __init__(self, name: str,
+                 weight: float,
+                 max_load_weight: int,
+                 current_load: Cargo = None,
```
Vorongor: You set current_load None by default, but it can also be a Cargo instance if it exists, here is a mismatch with annotation of type, it should be:  current_load: Cargo | None = None,

app/main.py
```diff
@@ -2,4 +2,64 @@ class Cargo:
...
+                 coords: list = None) -> None:
+
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+        if current_load:
+            self.hook_load(current_load)
```
Vorongor, line 58: But you've already set Cargo to current_load, that verification has no sense. You should add the unset option into hook method if Cargo too havy or verify it before set to current_load

Vorongor (approved):
Overall, the class hierarchy is structured well, demonstrating clear inheritance from BaseRobot → FlyingRobot → DeliveryDrone. The movement logic and base functionality are implemented correctly, and the design follows a logical OOP approach. You have a strong foundation here. With small refinements to type annotations and initialization logic, the code will be cleaner, more predictable, and easier to maintain.

PR #1866, opened 2025-11-01
Solution

leemonch1k-space:


app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
...
         self.weight = weight
 
-# write your code here
+
+class BaseRobot:
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
```
AngryAndrii: You wrote coords must be "list" type, but then appropriated default value as "None". Please, write a coorect type. You can read on "checklist" about that.

app/main.py
```diff
@@ -2,4 +2,72 @@ class Cargo:
...
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            coords: list = None
```
AngryAndrii: The same, please correct annotation type.

AngryAndrii (changes requested):

AngryAndrii (approved):

PR #1872, opened 2025-11-02
Soullution

Iventyk:


app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
     def __init__(self, weight: int) -> None:
         self.weight = weight
 
-# write your code here
+
+class BaseRobot():
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
```
vakt159, line 17: coords can be list and None so type hint should be updated

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+
+    def get_info(self) -> str:
+        return f"Robot: {self.name}, Weight: {self.weight}"
+
+
+class FlyingRobot(BaseRobot):
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        if coords is None:
+            coords = [0, 0, 0]
+        super().__init__(name, weight, coords)
```
vakt159: if coords is None:
            coords = [0, 0, 0]
can be replaced with coords = coords or [0,0,0]

app/main.py
```diff
@@ -2,4 +2,63 @@ class Cargo:
...
+            max_load_weight: int = 0,
+            current_load: Cargo = None
+    ) -> None:
+
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load = None
+
+        if current_load is not None:
+            self.hook_load(current_load)
```
vakt159, line 71: if current_load is not None:
        can replace with if current_load

vakt159 (changes requested):
Everything is good, requesting minor changes in type hint and some if statements

PR #1873, opened 2025-11-02
Solution

dmytro-malyk-dm:


app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+
+class FlyingRobot(BaseRobot):
+    def __init__(self,
+                 name: str,
+                 weight: int,
+                 coords: list[int] = None) -> None:
+        three_d_coords = coords or [0, 0, 0]
+        super().__init__(name=name, weight=weight, coords=three_d_coords)
+        self.coords = three_d_coords
+
```
AlexMishKo, line 45: using super() method you've asined the coords. no need of self.coords = three_d_coords

app/main.py
```diff
@@ -2,4 +2,68 @@ class Cargo:
...
+    def go_up(self, step: int = 1) -> None:
```
AlexMishKo, line 46: Using your code, it is possible to pass two-dimensional coordinates. This will cause an IndexError. Fix this, for example, by checking len() and using the .append method.

AlexMishKo (changes requested):

AlexMishKo (approved):

PR #1877, opened 2025-11-03
Solution

Guillelmina8:


tigersaymon (commented):
It's possible to rewrite the initialisation of the coordinates in fewer lines. Everything else is OK

tigersaymon (approved):

PR #2046, opened 2026-02-09
Solution: implemented classes to complete this task

SerhiiHryniuk:


app/main.py
```diff
@@ -2,4 +2,57 @@ class Cargo:
...
+
+class DeliveryDrone(FlyingRobot):
+    def __init__(
+            self,
+            name: str,
+            weight: int,
+            max_load_weight: int,
+            coords: list = None,
+            current_load: Cargo = None
+    ) -> None:
```
Reverad, line 62: here's wrong variables order.

Reverad (changes requested):
u could import annotations from __future__, for old python versions.

app/main.py
SerhiiHryniuk: I updated the parameter order because Python throws a SyntaxError: non-default argument follows default argument. In the previous version, max_load_weight (required) was placed after coords (which has a default value). I've moved all required parameters to the front to comply with Python's argument ordering rules.

image

I can add default parameter to max_load_weight as 0, but then it will be a little change for the description in the task, where does not say that max_load_weight must have a default parameter

Reverad: you right it's a bit off a task, but your max_load_weight can't be lower than 0 and in this way, ur saving the parameters order, which is important

Reverad (changes requested):
check my last comment and change parameter order pls

Reverad (approved):
nice, that's perfect

PR #2076, opened 2026-02-24
Solution

kjakub01:


kjakub01:
Can you point out the errors in the current code for me?

PR #2111, opened 2026-03-12
Solution

AlinaSv88:


app/main.py
```diff
@@ -2,4 +2,67 @@ class Cargo:
...
+        self.max_load_weight = max_load_weight
+        self.current_load = current_load
+        if isinstance(current_load, Cargo):
+            self.hook_load(current_load)
+
+    def hook_load(self, cargo: Cargo) -> None:
+        if isinstance(cargo, Cargo):
+            if self.current_load is None:
+                if cargo.weight <= self.max_load_weight:
+                    self.current_load = cargo
```
Boyok1337: This code could be improved by reducing nested if statements in hook_load() with early returns. It would also help to extract validation logic into a separate method, which would make the class cleaner and easier to maintain.

Boyok1337 (changes requested):

app/main.py
AlinaSv88: Done

app/main.py
```diff
@@ -2,4 +2,66 @@ class Cargo:
...
-# write your code here
+
+class BaseRobot:
+    def __init__(self, name: str, weight: int, coords: list = None) -> None:
+        self.name = name
+        self.weight = weight
+        if coords is None:
+            self.coords = [0, 0]
+        else:
+            self.coords = coords
```
vsmutok:
```suggestion
        self.coords = coords or [0, 0]
```
You can write it like this — it will look a bit more compact and cleaner. 🙂

vsmutok (changes requested):

app/main.py
AlinaSv88: Done

PR #2164, opened 2026-04-04
Solution

dimaaa-cmd:


app/main.py
```diff
@@ -2,4 +2,87 @@ class Cargo:
...
+        weight: int,
+        coords: list[int] | None = None,
+        max_load_weight: int = 0,
+        current_load: Cargo | None = None
+    ) -> None:
+        super().__init__(name, weight, coords)
+        self.max_load_weight = max_load_weight
+        self.current_load: Cargo | None = None
+
+        if current_load is not None:
```
PavelPereverzev1, line 77: It turns out current_load could be anything. It would make more sense to add a check like:
if isinstance(current_load, Cargo)
Because the function requires an instance of this type.

PavelPereverzev1 (approved):

PR #2206, opened 2026-04-22
Solution

RWAdelP:


RWAdelP:
@mateacademy-ai-mentor 
The review here made no sense! It kept specifying I need to handle2D vs 3D correctly 
If 2 then send the coords, if 3D then send the coords. That is VERY confusing feedback

PR #2271, opened 2026-06-14
Solution

Teashrock:


app/main.py
```diff
@@ -2,4 +2,74 @@ class Cargo:
...
+        weight: int,
+        coords: list[int] | None = None
+    ) -> None:
+        self.name = name
+        self.weight = weight
+        if not coords:
+            if self.__class__.__name__ == "BaseRobot":
+                self.coords = [0, 0]
+            else:
+                self.coords = [0, 0, 0]
```
vsmutok, line 19:
```suggestion
          self.coords =  coords or [0, 0]
```
You can write this a bit more simply.

vsmutok (changes requested):

PR #2339, opened 2026-08-11
Solution

sofiia-pohuliailo388:


Aragorn-untold (approved):
coords type hint should be list[int] not list[Any]

